PR #39, opened 2022-04-11
Use black

fscottfoti:
This PR is the result of running [black](https://github.com/psf/black) in the python directory (only reformatting is included, no code has changed).  

We could also:
* make black a dependency
* add a script to run black and prettier (for the JS parts) at the top level
* add a test to make sure each new PR has run the linting script

But I'm not sure any of that is necessary right now.

### DIFF
--- a/python/bin/pmtiles-convert
+++ b/python/bin/pmtiles-convert
@@ -1,18 +1,26 @@
 #!/usr/bin/env python
 
-#pmtiles to files
+# pmtiles to files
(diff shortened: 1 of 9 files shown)


bdon:
Thanks! this is a big help in making the project easier to contribute to.

fscottfoti:
No problem.  I'm a huge fan of autoformatting.  Also worth pointing out I didn't change any of the default settings.

Issue #38, opened 2022-04-08
Using python Reader object with cloud storage


fscottfoti:
It seems like `go-pmtiles` is set up to read from cloud storage (S3, Google Storage etc), but the python version here is not.  Is that correct?  How hard would it be to add that?  I use Google, but an S3 example would be great.

If the Reader API is robust, I can add it into a flask/fastapi endpoint pretty trivially, implement my own authentication etc, but it definitely seems like being able to read the tiles from a file in Storage (rather than locally) is the really powerful use case here.

bdon:
Yep, we should refactor the current python reader to take an object that implements the transport part.
Implementations that come to mind:
* mmap (current)
* http fetch via python standard urllib library
* s3, should have simple glue to boto interface

Open questions:
* should we use async/await? I would prefer to keep this as simple as possible

fscottfoti:
I would think async/await would not be necessary - is there an advantage to making multiple io requests at the same time, e.g. for the metadata and the file?  Seems like one has to wait on the other.

fscottfoti:
It seems like all you need to do is pass in a get_bytes function and it would support anywhere the user gets the data.  Do you need anything but the one function?  You could keep passing in a file as an option and build a default get_bytes function for that case that uses mmap.  The reader probably needs a test first for that kind of refactoring though.

fscottfoti:
This [PR](https://github.com/mapcraftlabs/PMTiles/pull/2/files) implements a get_bytes function as one way to do this.  I haven't tested it with Google Storage yet, but it seems to work with the local file.  I also have another PR I could open which reformats the python code with [black](https://github.com/psf/black) (like prettier in JS) [here](https://github.com/mapcraftlabs/PMTiles/pull/1).  Anyway it's just a draft for now - it would need some tests before merging.

fscottfoti:
OK I got it working with GCS.  Only hard part was I had to [refactor](https://github.com/mapcraftlabs/PMTiles/pull/2/files#diff-14e58dacda982cfdbd3bd578e05e3c7ad9779a11ed56559869b6116a3054c545R62-R73) load_directory so that there weren't so many separate calls for bytes from the server, but that ended up not being a big deal either and then it "just worked."  Very cool idea - takes about .2 seconds to fetch a tile in GCS, but it's probably even faster then called from an app deployed in the same data center.

bdon:
[...]

PR #42, opened 2022-04-26
refactor python reader to take a get_bytes function; add reader tests…

bdon:
… [#38]

Still todo:
- [ ] refactor reader to use less get_bytes calls, should only make one get_bytes for the header
- [ ] add other Source implementations

### DIFF
--- a/python/bin/pmtiles-show
+++ b/python/bin/pmtiles-show
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 import sys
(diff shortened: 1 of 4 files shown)


fscottfoti:
Interesting, yes, seems like a nice API.  My one comment was the first TODO item above so I think you've got it covered.  I'm still happy to do the Google Storage version once you have an AWS example for how to set up tests etc.

PR #50, opened 2022-06-23
Improve pmtiles-serve CLI description

flother:
Quick fix for the CLI description for `pmtiles-serve`. It looks like it was copy-and-pasted from [`pmtiles-convert`](https://github.com/protomaps/PMTiles/blob/cbc5f190838acafb0c2aca0450fef053eea977aa/python/bin/pmtiles-convert#L11), so it didn't describe the command correctly.


bdon:
Thanks!

PR #65, opened 2022-08-13
Implemented most of an attribute-matching parser

migurski:
Attribute parsing intended for selecting specific layers and attributes out of tiles, e.g.:

* `0/0/0.pbf` – include all attributes
* `0/0/0.pbf?attributes=*` – include all attributes
* `0/0/0.pbf?attributes=` – include **no** attributes
* `0/0/0.pbf?attributes=a1,a2` – Only attributes `a1` and `a2` from any layer
* `0/0/0.pbf?attributes=L1:a1,L2:b2` – Only attribute `a1` from layer `L1` and attribute `a2` from layer `L2`
* `0/0/0.pbf?attributes=L1:a1,L2:*` – Only attribute `a1` from layer `L1` and all attributes from layer `L2`
* `0/0/0.pbf?attributes=a1,L2:*` – Attribute `a1` from any layer and all attributes from layer `L2`

The only case I’m not handling reliably is backslashes in certain situations, just don’t think I’m doing the full recursive escape/parse dance correctly.

### DIFF
--- a/serverless/aws/test_util.py
+++ b/serverless/aws/test_util.py
@@ -1,5 +1,119 @@
 import unittest
+import util
 from util import parse_tile_path, pmtiles_path
+
+
+class TestAttributes(unittest.TestCase):
+    def test_just_attributes(self):
+        self.assertEqual(
+            util.parse_attributes("foobar"),
+            [(True, "foobar")],
+            "Should not split on missing comma",
+        )
+        self.assertEqual(
+            util.parse_attributes("foo,bar,baz"),
+            [(True, "foo"), (True, "bar"), (True, "baz")],
+            "Should split on simple comma",
+        )
+        self.assertEqual(
+            util.parse_attributes("foo\\,bar"),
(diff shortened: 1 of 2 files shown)


bdon:
@migurski OK, I think I've wrapped my head around the design here: wondering if there's a need for:

`0/0/0.pbf?attributes=L1:a1,L1:a2` - select multiple attributes in a single layer?

In the case where I select `a1` and `a2` from L1, do I then omit all features that have neither `a1` or `a2` or include them with no attributes? 

Is the intention of a top level `0/0/0.pbf?attributes=a1,a2` to include common keys like `name`? I suppose just specifying a big list of wanted attributes irrespective of layers is fine even if the named attribute only applies to a single layer.

bdon:
first pass of tile filtering implemented as a separate TS module in https://github.com/protomaps/PMTiles/compare/master...vtfilter but this isn't hooked up to any query string API yet.

PR #75, opened 2022-09-21
avoid allocation in rotate

missinglink:
can this trick be used to avoid allocating `t`?
https://go.dev/play/p/SMPKQaq-DDD


bdon:
yep, thanks! fixed in other repo (the spec-v3 here is a throwaway branch that will be deleted soon) 

https://github.com/protomaps/go-pmtiles/commit/9f750a88a843f8ed86c526ec55782d2ec5447d88

PR #76, opened 2022-09-21
code style: variable declarations

missinglink:
code style question, totally down to personal preference, and I was reluctant to mention it because of that...
seems that there's an opportunity to make the code a little easier to read without sacrificing correctness by using a more terse style of variable initialisation?

### DIFF
--- a/spec/converter/tile_id.go
+++ b/spec/converter/tile_id.go
@@ -15,13 +15,10 @@ func rotate(n int64, x *int64, y *int64, rx int64, ry int64) {
 }
 
 func t_on_level(z int64, pos int64) (int64, int64, int64) {
-	var n int64
-	n = 1 << z
+	var n int64 = 1 << z
 	rx, ry, t := pos, pos, pos
 	var tx int64
 	var ty int64
-	tx = 0
-	ty = 0
 	var s int64
 	for s = 1; s < n; s *= 2 {
 		rx = 1 & (t / 2)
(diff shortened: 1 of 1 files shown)


spec/converter/tile_id.go
```diff
@@ -120,16 +103,11 @@ func ZxyToQuadkey(z int64, px int64, py int64) int64 {
...
-	var y int64
-
-	var mask int64
-	mask = ^(0b1 << (z * 2))
+	var z = 32 - (bits.LeadingZeros64(uint64(i)) + 1) / 2
+	var mask int64 = ^(0b1 << (z * 2))
 
-	x = (i & mask)
-	y = (i & mask) >> 1
+	var x int64 x = (i & mask)
```
missinglink, line 109: oops

spec/converter/tile_id.go
```diff
@@ -36,23 +33,16 @@ func t_on_level(z int64, pos int64) (int64, int64, int64) {
...
 	var d int64
-	rx = 0
-	ry = 0
-	d = 0
-	var tx int64
-	var ty int64
-	tx = x
-	ty = y
+	var tx int64 = x
+	var ty int64 = y
```
missinglink, line 45: these guys probably don't even need a type defined

bdon:
cleaned up in other repo: https://github.com/protomaps/go-pmtiles/commit/a4ec835c501947eaf13f573541ed4679e0dec99f

PR #102, opened 2022-12-04
FetchSource handles servers that return 416 for requests beyond end of file

bdon:
* Refine the detection of servers that don't support byte serving.
* Improve console errors related to failed Fetch.


msbarry:
Looks good to me!

PR #112, opened 2023-01-11
js: FetchSource takes optional custom headers

bdon:
Allow passing of custom headers in FetchSource.

### DIFF
--- a/js/index.ts
+++ b/js/index.ts
@@ -293,7 +293,7 @@ export class FetchSource implements Source {
     return this.url;
   }
 
-  setHeaders(customHeaders:Headers) {
+  setHeaders(customHeaders: Headers) {
     this.customHeaders = customHeaders;
   }
 
--- a/js/index.ts
+++ b/js/index.ts
@@ -282,13 +282,19 @@ export class FileAPISource implements Source {
 
 export class FetchSource implements Source {
   url: string;
+  customHeaders: Headers;
 
-  constructor(url: string) {
+  constructor(url: string, customHeaders: Headers = new Headers()) {
     this.url = url;
+    this.customHeaders = customHeaders;
   }
 
   getKey() {
     return this.url;
+  }
+
+  setHeaders(customHeaders:Headers) {
+    this.customHeaders = customHeaders;
   }
 
   async getBytes(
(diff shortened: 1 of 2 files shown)


m-mohr:
This would be very helpful to pass authentication details.

bdon:
@m-mohr I should probably change the public API of `ol-pmtiles` to take an instance of `PMTiles` instead of a raw URL, letting you provide a custom `FetchSource` with headers and/or query params.

js/index.ts
```diff
@@ -292,9 +294,17 @@ export class FetchSource implements Source {
...
+		requestHeaders.set(
+			"Range",
+			"bytes=" + offset + "-" + (offset + length - 1)
+		);
+
 		let resp = await fetch(this.url, {
 			signal: signal,
-			headers: { Range: "bytes=" + offset + "-" + (offset + length - 1) },
+			headers: requestHeaders,
 		});
```
m-mohr, line 308: Would it makse sense to do something like this?

```js
		let options = {
			signal: signal,
			headers: { Range: "bytes=" + offset + "-" + (offset + length - 1) },
			headers: requestHeaders,
		};
		if (typeof this.prepareOptions === 'function') {
			options = this.prepareOptions(options, this.url);
		}
		let resp = await fetch(this.url, options);
```

this.prepareOptions is a callback that can be passed from the user and I assume the URL is passed by the user with potential query parameters anyway.

Then you have full flexibility.

Alternatively, you leave the full request to users. Then they can even choose the request library, e.g. axios.

bdon: custom FetchSource feels more flexible, I would want the default FetchSource to be quite narrow in its capabilities, maybe just adding header or query param for well-known use cases like auth,

The big feature missing right now in FetchSource and what will demand a reworking of the `Source` API is support for `If-Match` ETag headers.

m-mohr: Can this interface maybe just look as the interface for https://openlayers.org/en/latest/apidoc/module-ol_source_GeoTIFF.html#~GeoTIFF looks like? There you have a headers field in sourceOptions.

bdon: @m-mohr I think having `headers` of type `Object.<string, string> | undefined` is good, as long as the headers is static for the lifetime of the Source object. Is this true in your authentication cases, or does it require some dynamic token to be passed that might have been refreshed in a single session?

m-mohr: That should be fine, I think.

bdon: We can make `ol-pmtiles` conform to the conventions of the other sourceOptions then - I'll plan to work on this during the STAC sprint

bdon: Merging a basic additional param here that takes a `Headers` object. We can make it look more OL-friendly in `ol-pmtiles`

PR #117, opened 2023-01-28
Fix bug in find_tiles in the python V3 code

fscottfoti:
The good news is this stuff is pretty obviously broken when it's broken. I noticed that the JS example was working with my file but the python version was not (most tiles were working, but some were not), so I traced the differences until I found the problem. Let me know if this looks right.

### DIFF
--- a/python/pmtiles/v2.py
+++ b/python/pmtiles/v2.py
@@ -1,3 +1,4 @@
+import json
 from collections import namedtuple
 
 
--- a/python/pmtiles/tile.py
+++ b/python/pmtiles/tile.py
@@ -85,7 +85,7 @@ def tileid_to_zxy(tile_id):
 def find_tile(entries, tile_id):
     m = 0
     n = len(entries) - 1
-    while m < n:
+    while m <= n:
         k = (n + m) >> 1
         c = tile_id - entries[k].tile_id
         if c > 0:


python/pmtiles/v2.py
```diff
@@ -1,3 +1,4 @@
+import json
```
fscottfoti, line 1: Unrelated, but also necessary

bdon: @fscottfoti the JS version is backwards compatible with v2; do you have a need for this for the python client as well, instead of converting your archives from v2 to v3 with https://github.com/protomaps/go-pmtiles ?

For some context, I don't have an internal use case for the Python library ATM since the Lambda implementation at https://github.com/protomaps/PMTiles/tree/main/serverless/aws changed from Python to Node for v3. So if there is 0 use cases for v2 reading in Python I'd prefer to remove this module completely.

bdon:
Thanks! Will review this and improve the tests later tonight

python/pmtiles/v2.py
fscottfoti: Not really a use case but it was handy to be able to upgrade *before* upgrading all the files.  I was appreciative it was there.  If you decide to keep it you could always keep the single most recent version or something but it's not required.

Thanks for keeping the python API without a personal use case.  It's much appreciated.

bdon:
Thanks for the patch!

Issue #111, opened 2023-01-05
Unclear Metadata (JSON?) spec


nyurik:
The v2 of the spec mentions that metadata must be a proper json:

> Followed by M bytes of metadata, which must be a JSON string with bounds, minzoom and maxzoom properties (new in v2)

In v3, it seems all requirements for metadata are gone, with some fields (bounds, zooms, etc) are now part of the fixed header.  Are there any specs, or even just recommendations for consistency of the fields that are still part of the JSON? There seem to be some expectations in [this code](https://github.com/protomaps/PMTiles/blob/af409af0437b20ff4a9e3ba95004392c023edbb9/app/src/MaplibreMap.tsx#L101) of `tilestats` and `vector_layers` -- the later is probably part of tilejson spec, but what is the `tilestats`?

Lastly, unrelated - could you rename the `spec.md` to `README.md` in those dirs? makes browsing a bit easier :)

bdon:
the json metadata is intended to be application specific. I'm probably going to add `vector_layers` as defined by https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#vector_layers but maybe limited to just `id` instead of having to enumerate all fields (what about tiles where the # of tags is not a fixed tabular schema, for example?) but haven't formalized it out of caution for unexpected but common use cases.  

`tilestats` is something generated by Tippecanoe already which includes summary statistics. This is necessary if you want to style your map based on min and max like a choropleth.

bdon:
Current things under consideration:

* `type:baselayer`if the raster or vector tileset is a basemap; if this is NOT true, like `type:overlay`, a client can visualize the tileset over a basemap
* `vector_layers:` generally following MBTiles/TileJSON but looser language to allow for describing OSM QA Tiles https://osmlab.github.io/osm-qa-tiles/
   *Maybe adding another value indicating all of the geometries in a vector tile layer are the same geometry type? MVT allows heterogeneous layers, but in practice many systems emit only one geometry type per layer.

bdon:
@nokonoko1203 @nyurik @msbarry this is the minimal refinement to the spec I'm going propose: https://github.com/protomaps/PMTiles/pull/123

Note that http://github.com/felt/tippecanoe already behaves exactly this way.

How it affects other programs:

Program to convert from MBTiles to PMTiles needs to check for the existence of a `json` key in metadata, and extract the two sub-keys `vector_layers` and `tilestats` into separate keys

[...]

PR #123, opened 2023-02-20
add spec 3.1 metadata details [#111]

bdon:


bdon:
Updated with:

* Addition of AVIF tiletype
* Clarify `vector_layers` requirement for MVT filetypes to match [TileJSON 3.0](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0#33-vector_layers) - (it is not worth innovating here and backwards compatibility is more important)
* Clarify requirements around non-zero directory and tile-lengths

nyurik:
thx for the update!  could you rename them to readme for ease-of-consumption?  requires an extra click plus minor confusion otherwise

bdon:
@nyurik I can survive with the extra click, there are existing links to that GitHub path, plus we might want to use `README.md` for meta-information about the spec (describing test datasets, etc) in the future

nyurik:
would it work if you leave the link files to the readme?  Testing could be listed as part of the spec in a separate section, doesn't have to be THAT formal :)

bdon:
@nyurik I meant there are external links to the URL https://github.com/protomaps/PMTiles/blob/main/spec/v3/spec.md

nyurik:
@bdon I understand, I was just wondering if you leave a `spec.md` as a symlink to README, if it will work? Or if you create a README.md as a symlink to the spec.md, if github would show it at the bottom of the page

bdon:
@nyurik I am trying to minimize the surface area of files/symlinks/repo artifacts I need to maintain. Is there an alternate solution to an extra click? Given that other formats such as [COG](https://github.com/cogeotiff/cog-spec/blob/master/spec.md) and [MBTiles](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md) name their specs `spec.md` on GitHub I would prefer to keep it as-is.

PR #129, opened 2023-03-06
clarify build instructions for AWS lambda function

jamesscottbrown:
Doing a clean checkout then npm run build isn't enough to build the Lambda function - it will result in errors like 

```
X [ERROR] Could not resolve "fflate"

    ../../js/index.ts:1:31:
      1 │ import { decompressSync } from "fflate";
        ╵                                ~~~~~~~~

  You can mark the path "fflate" as external to exclude it from the bundle, which will remove this
  error.
```


bdon:
Thanks!

PR #131, opened 2023-03-10
disable "Load URL" button if url is empty

jamesscottbrown:
This prevents the user from accidentally clicking the button if they try to quickly click on the file drop-zone (currently, doing this would display an error, and the user would need to press the back button in their browser to return to the start form).


bdon:
Thanks!

PR #132, opened 2023-03-11
Add prettier as dev dependency with config file

hu0p:
I was doing some optimization work in `js/` and noticed that the diff for the PR I'm about to introduce is difficult to read due to differing local configuration on my part.

This PR introduces the same (default) prettier setup as [protomaps.js](https://github.com/protomaps/protomaps.js) to `js/` in order to avoid confusion for the maintainer(s) and contributors.

I added prettier as a dev dependency, added a `default prettierrc.json` to prevent common editor plugin configuration from being used instead, and ran prettier within the `js/` project folder.

Stay tuned for the other PR coming momentarily!


bdon:
Thanks! I'll go through the repo and make the formatting consistent with the other files too.

PR #136, opened 2023-03-15
Add WebP PMTiles binary & example file

roblabs:
- [ ] When this PR lands, then the URL to the WebP PMTiles can properly be tested.
  * committed path is `spec/v3/usgs-mt-whitney-8-15-webp-512.pmtiles`, and the GitHub Action will make use of it.

---

*Added a WebP PMTiles sample file (2 MB)*


roblabs:
I am not sure what the [CI:build](https://github.com/protomaps/PMTiles/actions/runs/4422560715/jobs/7754462958#step:6:20) failure is due to.  I saw this on local builds, but I'm not certain what part of prettier isn't working.  Please reject and I'll rework as needed.

bdon:
I fixed it by running `npm run prettier` in `app/` (it just wanted a comma at the end of the last array element)

Thanks for contributing the sample!

Issue #3, opened 2021-02-24
OpenLayers plugin


bdon:

fl0cke:
I just discovered this project and official support for Openlayers would be awesome 👍 Btw here is an open issue that discusses support for alternative tile sources like yours: https://github.com/openlayers/openlayers/pull/12008#issue-570057745 So maybe you could contribute to those efforts?!

bdon:
Hi, agreed this would be great! The major blocker is for the OpenLayers API to allow asynchronous tile creation, which I was not able to figure out based on looking at the code or any plugins. If anyone has any tips or can show an example I would be happy to continue on this path.

fl0cke:
Hey, here is the code i'm currently using:

```
import { TileImage } from 'ol/source'
import State from 'ol/source/State'

import { extentFromProjection } from 'ol/tilegrid'
import TileGrid from 'ol/tilegrid/TileGrid'
import { getHeight, getWidth } from 'ol/extent'
import TileState from 'ol/TileState'

function shift(number, shift) {
  return number * Math.pow(2, shift)
}

function getUint24(dataview, pos) {
  return (
    shift(dataview.getUint16(pos + 1, true), 8) + dataview.getUint8(pos, true)
  )
}

function getUint48(dataview, pos) {
  return (
    shift(dataview.getUint32(pos + 2, true), 16) + dataview.getUint16(pos, true)
  )
}

export default class PMTiles extends TileImage {
  constructor(options) {
    const extent = options.extent || extentFromProjection(options.projection)

    const tierSizeInTiles = []
    const width = getWidth(extent)
    const height = getHeight(extent)
    const tileSize = options.tileSize
    let tileSizeInTier = tileSize

    while (width > tileSizeInTier || height > tileSizeInTier) {
      tierSizeInTiles.push([
        Math.ceil(width / tileSizeInTier),
        Math.ceil(height / tileSizeInTier),
      ])
      tileSizeInTier += tileSizeInTier
    }

    tierSizeInTiles.push([1, 1])
    tierSizeInTiles.reverse()

    const resolutions = [1]
    for (let i = 1, ii = tierSizeInTiles.length; i < ii; i++) {
      resolutions.push(1 << i)
    }
    resolutions.reverse()

    const tileGrid = new TileGrid({
      tileSize,
      extent,
      resolutions,
    })

    super({
      state: State.LOADING,
      attributions: options.attributions,
      cacheSize: options.cacheSize,
[...]

PR #142, opened 2023-03-22
Fix tile loading

weskamm:
When testing the openlayers demo application, we discovered an issue after a certain amount of features had been loaded where no new tiles will get fetched anymore.

To see the issue in action one needs to zoom in and out a lot on different locations to fetch a lot of features.
After some time, ~1 minute, tiles will stop loading and the map gets blurry when zooming in and does not display new features.

It seems related to the following issue https://github.com/openlayers/openlayers/issues/8276
which inspired the change made here.
After that change, tiles load correctly.

### DIFF
--- a/js/examples/openlayers.html
+++ b/js/examples/openlayers.html
@@ -46,6 +46,7 @@
                       tile.setFeatures(features);
                       tile.setState(2); // LOADED
                     } else {
+                      tile.setFeatures([]);
                       tile.setState(4); // EMPTY
                     }
                   });


bdon:
Thanks for the patch! Going to verify in https://github.com/protomaps/PMTiles/issues/3 before merging

bdon:
@weskamm I'm having trouble reproducing the original issue of tiles stopping loading - are you zooming in/out on blank areas or areas with lots of features? can you also verify you are setting `maxZoom` correctly to match your PMTiles file in the `VectorTile` source, instead of what it's set to in the example (14) ?

weskamm:
hi, i could reproduce the issue with the current example as is, without modifications.
As said, i had to fill up the cache with lots of features by zooming in and out on different areas where a lot of features are.
After 1, max. 2 minutes, the problem got visible.

bdon:
@weskamm thanks - also what browser/platform (mobile, desktop) were you reproducing this with?

weskamm:
chromium 109.0.5414.74 on debian linux. just tried again, example broke again after ~ 1min. of pan / zoom

bdon:
OK, I was able to reproduce it, after merge couldn't get it to happen. Thanks for the patch!

PR #146, opened 2023-03-27
PMTiles Viewer invalid link href in GitHub Pages - Bugfix

kajkal:
Clicking the link in the title (open in new tab/window) results in opening https://protomaps.github.io/ instead of https://protomaps.github.io/PMTiles/.
Which is odd considering that the base url is set to `/PMTiles/` in [the build script](https://github.com/protomaps/PMTiles/blob/081be12fdbbc33eda7c41f203063b3c596eed719/.github/workflows/actions.yml#L23).

Setting `href={import.meta.env.BASE_URL}`, from my local attempts, doesn't look like it can fix the error.
Thus this straightforward fix below.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -101,7 +101,7 @@ function App() {
   return (
     <div>
       <Header>
-        <Title href="/" onClick={clear}>
+        <Title href="/PMTiles/" onClick={clear}>
           PMTiles Viewer
         </Title>
         <GithubLink>


bdon:
Thanks! Seems like a mistake on my side that the title link is used to clear the selected archive and doesn't actually work like a normal link - how can we improve this?

kajkal:
This clear archive on click feature is not a problem. When I converted this title to a link, I just didn't expect that it wouldn't work properly deployed on GitHub Pages. Now it works properly, clears the current archive and is able to quickly open new tabs/windows with the viewer application. 
In my opinion, the way it is now is fine 👍

PR #152, opened 2023-04-04
Delete unused definition of entry_zxy_cmp

jamesscottbrown:
See https://github.com/felt/tippecanoe/pull/91


bdon:
Thanks! this can be copied into tippecanoe now

PR #154, opened 2023-04-07
set blob type for leafletRasterLayer

bmcbride:
Setting the mime type helps with previewing network requests.


bdon:
Nice enhancement, thanks!

PR #172, opened 2023-05-06
More detailed Version 3 Specification

### LINKED ISSUE #107: More detailed Specification
Hey :)

First of all. Thank you for the great format. It is imho the perfect solution to a problem that will become more and more prevalent in the near future.

I tried to implement my own reader / writer in rust and found myself looking at the first-party implementations quite a lot, because I stumbled upon things that are not really clear in the specification.

I think the project would benefit a lot from having a WAY MORE detailed specification (at least more than something I could print on a single page) and I was wondering whether a contribution on my part would be welcomed.

I took the opportunity and wrote [a really rough draft](https://gist.github.com/DerZade/cda63bccd24b3999b04e429955f61bfb), of the section on the header. Just to give you an impression of what I would envision and get your feedback.

PS: It's also totally fine if you do not think the specification needs to be improved.

DerZade:
This PR fixes #107, by attempting to introduce a more detailed specification for version 3.


spec/v3/spec.md
```diff
@@ -1,93 +1,372 @@
...
-	* Enforce a canonical extension e.g. `.mvt`, `png`, `jpeg`, `.webp` to prevent duplication in caches
-
-### Organization
-
-In most cases, the archive should be in the order `Header`, Root Directory, JSON Metadata, Leaf Directories, Tile Data. It is possible to relocate sections other than `Header` arbitrarily, but no current writers/readers take advantage of this. A future design may allow for reverse-ordered archives to enable single-pass writing.
+# PMTiles Version 3 Specification
+
+## 1 Introduction
+
+PMTiles is a single-file archive of square tiles.
```
DerZade, line 5: This still needs more work. I would love to include a very brief "Why PMTiles?"

spec/v3/spec.md
```diff
@@ -1,93 +1,372 @@
...
+
+## 2 Overview
+
+A archive consist of five main sections:
+
+1. A fixed-size 127-byte header (described in [chapter 3](#3-header))
+1. A root directory (described in [chapter 4](#4-directories))
+1. Optional JSON metadata (described in [chapter 5](#5-json-metadata))
```
DerZade, line 13: @bdon in https://github.com/protomaps/PMTiles/issues/107#issuecomment-1535056512:

> I think we should clarify the spec to make JSON required: the null case is just the empty JSON object. I don't see any downside to doing this and it simplifies the client.

spec/v3/spec.md
```diff
@@ -1,93 +1,372 @@
...
+
+#### Metadata Offset
+
+The Metadata Offset is a 8-byte field whose value gives the offset of the first byte of the metadata. This address offset is relative to the first byte of the archive.
+
+This field is encoded as an little-endian 64-bit unsigned integer.
+
+#### Metadata Length
+
+The Metadata Length is a 8-byte field specifying the number of bytes reserved for the metadata. A value `0` indicates that there is no metadata included in this PMTiles archive.
```
DerZade, line 75: @bdon in https://github.com/protomaps/PMTiles/issues/107#issuecomment-1535056512:

> I think this could better be "number of bytes used", since reserved might imply there is empty space in a typical file.

spec/v3/spec.md
```diff
@@ -1,93 +1,372 @@
...
+
+#### Examples
+|Tile-ID|Offset|Length|Run-Length|Description|
+|--:|--:|--:|--:|:--|
+|`5`|`1337`|`42`|`1`|Tile 5 is located at bytes 1337-1378 of the _tile data section_|
+|`5`|`1337`|`42`|`3`|Tile 5, 6 and 7 are located at bytes 1337-1378 of the _tile data section_|
+|`5`|`1337`|`42`|`0`|A leaf directory whose first tile has an ID of 5 is located at byte 1337-1378 of the _leaf directories section_|
+
+### 4.2 Encoding
+A directory can only be encoded in its entirety. It is not possible to encode a single directory entry by itself.
```
DerZade, line 280: @bdon in https://github.com/protomaps/PMTiles/issues/107#issuecomment-1535056512:

> What implies this? An archive that addresses a single tile should be valid, AFAIK.

spec/v3/spec.md
DerZade: Theoretically you could reserve more space for any section (apart from the meta data) than it actually needs 🤔. Or am I getting something wrong here? 🤔 For example: I could reserve 30MB for the tile data section, but only have 10MB of tile data.

spec/v3/spec.md
DerZade: Fine by me. I remember running into this when trying out my own encoder.
Can this be any valid JSON value or does this have to be a object? 42, null or an array would also be valid JSON 🤔I personally would welcome that. You could make a good case about at least supporting top level arrays and if you support arrays and objects you might as well allow any valid JSON 😅

spec/v3/spec.md
DerZade: The values you encode for an entry always depend on the entry that came before that entry. That's why I said that entries can't be encoded by itself.

Yes an archive that addresses a single tile should be valid, but that would have to have a root directory with one entry. At least you would to have to prepend a `1` to indicate the number of entries, that are contained in the directory and at that point it's a directory and not a single entry.

spec/v3/spec.md
DerZade: Addressed in [cdf746b](https://github.com/protomaps/PMTiles/pull/172/commits/cdf746b5bceb1d5a846087060592f3e91450ff18)

spec/v3/spec.md
bdon: > At least you would to have to prepend a 1 to indicate the number of entries, that are contained in the directory and at that point it's a directory and not a single entry.

OK, I see what you mean here. I meant to say that a directory with length 1 is valid; any directory with length > 0 is valid. So I think 4.2 should simply say "A directory must have at least one entry".

spec/v3/spec.md
```diff
@@ -1,110 +1,460 @@
...
+
+#### Metadata Offset
+
+The Metadata Offset is an 8-byte field whose value gives the offset of the first byte of the metadata. This address offset is relative to the first byte of the archive.
+
+This field is encoded as a little-endian 64-bit unsigned integer.
+
+#### Metadata Length
+
+The Metadata Length is an 8-byte field specifying the number of bytes reserved for the metadata.
```
bdon, line 90: Here and in other sections we should never use the word "reserved". Reserved implies that we might have allocated more bytes than is necessary. For example if there is 100 bytes of valid metadata, we never want there be 101 bytes, or 200 bytes. A parser should not be expected to deal with extra zero or irrelevant bytes past the end of the data; and extra bytes at the end don't serve any purpose in the current design as no affordances are provided for overwriting of data in-place.

spec/v3/spec.md
DerZade: I backported the changes from 3.1 in 60cb7259e, which requires it to be an object.

spec/v3/spec.md
DerZade: I'm happy to add something along the lines of that, although it's already mentioned in another chapter.

But I still think we should keep something along the lines of _"A directory can only be encoded in its entirety"_, because I remember that wasn't really clear to me when I first started implementing PMTiles. What I wanted to know then was how the entries are encoded / decoded, but that's just the wrong starting point, since they are always interleaved.

spec/v3/spec.md
DerZade: Continued [here](https://github.com/protomaps/PMTiles/pull/172#discussion_r1266232148)

spec/v3/spec.md
DerZade: Addressed in c3a6c6ffe

PR #176, opened 2023-05-15
Remove max-width for tooltips

jamesscottbrown:
I have some map tiles with long property names - currently the tooltip is too narrow, so text extends beyond the tooltip's white background.

The Maplibre docs say that `options.maxWidth` is "A string that sets the CSS property of the popup's maximum width, eg '300px' . To ensure the popup resizes to fit its content, set this property to 'none'"

https://maplibre.org/maplibre-gl-js-docs/api/markers/#popup


bdon:
Thanks for the enhancement!

PR #183, opened 2023-05-25
Fix tile-finding false positives in find_tile

e-n-f:


python/test/test_tile.py
```diff
@@ -106,11 +106,16 @@ def test_find_tile_multiple(self):
...
     def test_find_tile_leaf(self):
-        entries = [Entry(100, 1, 1, 0)]
+        entries = [Entry(100, 1, 1, 151)]
         result = find_tile(entries, 150)
         self.assertEqual(result.offset, 1)
         self.assertEqual(result.length, 1)
 
+    def test_find_tile_missing_between(self):
+        entries = [Entry(3, 3, 1, 0),Entry(5, 5, 1, 0)]
+        result = find_tile(entries, 4)
```
bdon, line 116: `find_tile(entries,4)` should return the first Entry(TileID=3) because `Entry(3,3,1,0)` is a leaf and might contain any tile data in the range `[3,5)` - perhaps we should rename it to `find_relevant_entry`?

bdon:
@e-n-f what specific missing tile was this happening on?

e-n-f:
I am traveling and don't have the details immediately at hand, but will add a comment once I do.

`find_tile(entries, 4)` should return `None` because `Entry(3,3,1,0)` has a `length` of 0 and therefore doesn't represent any duplicates, right?

bdon:
@e-n-f run lengths of `0` mean the pointed-to byte sequence represents not a tile but another [directory of entries](find_tile(entries, 4) should return None because Entry(3,3,1,0) has a length of 0 and therefore doesn't represent any duplicates, right? Or do I misunderstand the structure?). The case where there are no duplicates has a run length of 1

e-n-f:
I'm sorry, I can't seem to reproduce the problem we were seeing with a local pmtiles file. Perhaps the bug is actually in our `get_bytes` function? Are the offset and length that are passed in to the function ever out of bounds in a way that needs error-checking, or are they always valid? @migurski, do you have an example that is reproducible outside of that context?

bdon:
@e-n-f the problem I am aware of related to possibly OOB garbage entries being created at specific directory sizes should have been fixed in https://github.com/felt/tippecanoe/commit/e43d6dacdda556330759c13d60eda736cec45ba1 - is the problem file created with a build before that commit?

e-n-f:
The file in question was created as mbtiles by `gdal_translate` and converted to pmtiles with `pmtiles convert` rather than involving tippecanoe at all.

bdon:
@e-n-f this is specific to the python implementation right? it's much less production-tested at this point compared to the JS one so totally possible the `get_bytes` is erroneous... provide a sample file reproducing it if possible.

migurski:
Yes, we’re finding this in the Python implementation.

e-n-f:
I am no longer able to reproduce this problem and think that I must have misdiagnosed it.

bdon:
OK, for the record we saw some issues at the boundaries of zoom levels so there might still be something there; reopen if anyone discovers it

PR #185, opened 2023-05-28
reuse `Reader.header`

### LINKED ISSUE #179: [python] header is deserialized multiple times on read
The `deserialize_header` function is called in multiple methods - `Reader.header`, `Reader.metadata`, and `Reader.get`.  Each `Reader` class is logically associated with a single pmtiles file, which means the header could be deserialized once and reused for the lifetime of the `Reader` instance.

It makes sense to me to cache the deserialized header as an instance attribute of the `Reader` class instead of deserializing it multiple times.  If this behavior is not desired then it makes sense to at least call `self.header()` from the `metadata` and `get` methods instead of repeating the call to `deserialize_header` (DRY).

geospatial-jeff:
Closes #179

### DIFF
--- a/python/pmtiles/reader.py
+++ b/python/pmtiles/reader.py
@@ -35,7 +35,7 @@ class Reader:
         return deserialize_header(self.get_bytes(0, 127))
 
     def metadata(self):
-        header = deserialize_header(self.get_bytes(0, 127))
+        header = self.header()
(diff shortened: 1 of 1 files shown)


vincentsarago (approved):

bdon:
Thanks!

PR #186, opened 2023-05-28
deserialize_header returns pmtiles version

### LINKED ISSUE #180: [python] `deserialize_header` should return pmtiles version
The method checks that the pmtiles archive is version 3 but doesn't return the version.  I think `deserialize_header` should transparently return all key/values in the header, it shouldn't be stripping out any information.

```python
{"version": buf[7]}
```

I'm happy to open a PR!

geospatial-jeff:
Closes #180


bdon:
LGTM!

PR #191, opened 2023-06-07
serverless: improve error message if request doesn't match archive type

bdon:
See #190


MzicoDim:
This would be very helpful 👍

PR #192, opened 2023-06-09
Python: mbtiles_to_pmtiles(): patch maxzoom metadata when using maxzoom option

rouault:


bdon:
Nice catch! Thanks for the PR.

PR #193, opened 2023-06-09
Python: mbtiles_to_pmtiles(): set tile_compression = gzip when converting from MVT MBTiles

rouault:


python/pmtiles/convert.py
```diff
@@ -38,8 +38,8 @@ def mbtiles_to_header_json(mbtiles_metadata):
     else:
         header["tile_type"] = TileType.UNKNOWN
 
-    if mbtiles_metadata.get("compression") == "gzip":
-        header["tile_compression"] = Compression.GZIP  # TODO: does this ever matter?
+    if tile_format == "pbf" or mbtiles_metadata.get("compression") == "gzip":
```
bdon, line 41: I don't think this will correctly handle MBTiles MVT archives with uncompressed tiles. These are created via existing tools like:

`./tippecanoe ne_50m_admin_0_countries.geojsonseq --no-tile-compression -o out.mbtiles -z6 --force`

AFAIK the `compression` field has never been a formal part of the MBTiles spec: described in "Future Directions" at [spec 1.3](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#future-directions)

The MBTiles specification is also open ended for other formats: See [Spec 1.3 content](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#content). So I believe we must ignore content metadata and only inspect the magic number to determine whether the content is gzipped. I understand this isn't 100% bulletproof but in my experience should cover vector tile use cases as well as cases like gzipped raw pixel grids as tiles.

bdon: Nevermind, missed the context in which this works (ensures GZIP pmtiles output regardless of input)

PR #194, opened 2023-06-09
pmtiles.hpp: multiple warning and warning fixes

rouault:
This fixes a number of warnings found when integrating it in GDAL builds, which are done with quite pedantic warning levels.

- add explicit casts when shortening variables
- add support for big-endian targets if defining PMTILES_MSB
- add santity checks on reading directories against corrupt/hostile data
- replace a exit() call by an exception


cpp/pmtiles.hpp
```diff
@@ -165,11 +198,13 @@ struct entryv3 {
 	}
 };
 
+#ifdef PMTILES_NEED_ENTRYV3_CMP
```
bdon, line 201: Maybe we should just remove this struct if not all consumers need it ( the only significant open-source consumer I'm aware of is https://github.com/felt/tippecanoe/blob/main/pmtiles_file.cpp )

rouault: hum, for now, I'm only in read-support, but later I'll implement write support as well, and I'll probably need to sort entries too, so it might be useful. So I'm going to remove this #ifdef guard

bdon:
Fabulous! merged.

> This fixes a number of warnings found when integrating it in GDAL builds, which are done with quite pedantic warning levels.

Is there somewhere these warning levels are defined so we could reproduce them here to catch warnings in the future? 

> add santity checks on reading directories against corrupt/hostile data

FYI, for write support we may want to narrow the specification to add a limit on the byte size or # of entries in a leaf directly According to the current spec https://github.com/protomaps/PMTiles/blob/main/spec/v3/spec.md a single huge leaf directory is valid, but a 100MB+ leaf directory would cause problems on [runtimes like Cloudflare](https://github.com/protomaps/PMTiles/tree/main/serverless/aws) with a memory limit of 128MB. 

I'm only familiar with the writers authored mostly by me at [go-pmtiles](https://github.com/protomaps/go-pmtiles), [cpp/pmtiles.hpp](https://github.com/protomaps/PMTiles/blob/main/cpp/pmtiles.hpp) and [planetiler](https://github.com/onthegomap/planetiler/blob/main/planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java). None of these writers currently write more than 1 level of leaves. 

The potential issue with constraining to a limited size leaf is it will implicitly create a maximum addressable archive size for single-leaf-level archives. In practice, this has never happened since the use cases I'm aware of are z0-z15 tilesets with significant redundancy. Adding GDAL write support and/or attempting larger tilesets (z16, less repetitive tilesets like DEMs) is a good point to revisit this.

rouault:
> Is there somewhere these warning levels are defined so we could reproduce them here to catch warnings in the future?
> 
> >

https://github.com/OSGeo/gdal/blob/master/gdal.cmake#L133 to https://github.com/OSGeo/gdal/blob/master/gdal.cmake#L177 is warning flags that GDAL tries to add for gcc/clang
and https://github.com/OSGeo/gdal/blob/master/gdal.cmake#L72 for MSVC

bdon:
https://github.com/protomaps/PMTiles/pull/198

PR #212, opened 2023-06-30
Update tippecanoe URL

ay8s:
Looks like there was maybe a fork of [tippecanoe](https://github.com/mapbox/tippecanoe) previously which no longer exists. Adjusted the URL to the Mapbox repo.


nvkelso (commented):
Should this go to the Felt org repo instead? That's the maintained repo these days?

PR #231, opened 2023-08-06
add ETag and cache-control headers for 200 responses

blacha:
Using a sha256sum of the data seemed liked the easiest option for the ETag it also means that if the pmtiles file changes but the data does not the tile can be cached between updates.

### DIFF
--- a/serverless/aws/src/index.ts
+++ b/serverless/aws/src/index.ts
@@ -208,7 +208,7 @@ export const handlerRaw = async (
       }
 
       headers["Cache-Control"] = `public, max-age=${process.env.CACHE_MAX_AGE || 86400}`;
-      headers["eTag"] = `"${createHash("sha256").update(Buffer.from(data)).digest("hex")}"`
+      headers["ETag"] = `"${createHash("sha256").update(Buffer.from(data)).digest("hex")}"`
 
       if (is_api_gateway) {
         // this is wasted work, but we need to force API Gateway to interpret the Lambda response as binary
--- a/serverless/aws/src/index.ts
+++ b/serverless/aws/src/index.ts
@@ -14,6 +14,7 @@ import {
 import { pmtiles_path, tile_path, tileJSON } from "../../shared/index";
 
 import zlib from "zlib";
+import { createHash } from "crypto"
 
 import { S3Client, GetObjectCommand } from "@aws-sdk/client-s3";
 import { NodeHttpHandler } from "@aws-sdk/node-http-handler";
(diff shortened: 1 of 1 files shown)


serverless/aws/src/index.ts
```diff
@@ -206,6 +207,9 @@ export const handlerRaw = async (
         data = tilePostprocess(data, header.tileType);
       }
 
+      headers["Cache-Control"] = `public, max-age=${process.env.CACHE_MAX_AGE || 86400}`;
+      headers["eTag"] = `"${createHash("sha256").update(Buffer.from(data)).digest("hex")}"`
```
bdon, line 211: even though headers are treated as case-insensitive I think the well-known case is `ETag` right?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

bdon:
Thanks! If we eventually move to [streaming responses](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/) we would have to change this right? Because we wouldn't have the entire tile contents, we'd just return a DecompressionStream.

The only other options I see are using the archive etag + the tile coordinates, but that doesn't have the nice content-aware property.

serverless/aws/src/index.ts
blacha: Good spotting, have updated it now.

blacha:
> If we eventually move to [streaming responses](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/) we would have to change this right?

Yeah a different ETag would be needed, one option would be to somehow store the  content hash in the pmtiles archive and then serve that. 

I haven't seen any benchmarks with streaming responses, I was under the impression it mostly helps with larger files (2MB+) I am unsure how much you would gain by switching for smaller <256kb files. I would be very interested in any benchmarks if you have found any.

serverless/aws/src/index.ts
```diff
@@ -235,7 +239,6 @@ export const handlerRaw = async (
     }
     throw e;
   }
-  return apiResp(404, "Invalid URL", false, headers);
```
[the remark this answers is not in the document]
bdon: Yep, fine to remove

bdon:
@blacha BTW, one perspective on adding `Cache-Control` here is if the lambda is always going to be used with CloudFront then it makes more sense to let the configuration of Cache-Control live in CloudFront configuration instead of the code/environment variable. So maybe it doesn't belong here? The ETag is definitely useful.

bdon:
Nevermind - I was mistaken about the CloudFront behavior, it seems to require the Origin (Lambda) to provide the header, it doesn't add it itself.

PR #235, opened 2023-08-17
Spec v3 edits

bdon:
@DerZade making some edits to formatting and wording

### DIFF
--- a/spec/v3/spec.md
+++ b/spec/v3/spec.md
@@ -6,14 +6,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 Please refer to the [change log](./CHANGELOG.md) for a documentation of changes to this specification.
 
+## 1 Abstract
 
-## 1 Abstact
-
-PMTiles is a single-file archive format for tiled data. It enables low-cost, zero-maintenance map applications for "serverless" environments without having to rely on a custom tile backend or a third-party provider. This is achieved by packing all tiles of a tileset into an archive so that all tiles can be accessed easily and without much overhead via HTTP range requests. By combining all the tiles into one archive, hosting costs are kept low, as it is usually a lot cheaper to update one large file than to update thousands or even millions of small files.
+PMTiles is a single-file archive format for tiled data.
 
 ## 2 Overview
 
-An archive consists of five main sections:
+A PMTiles archive consists of five main sections:
 
 1. A fixed-size 127-byte header (described in [Chapter 3](#3-header))
 1. A root directory (described in [Chapter 4](#4-directories))
@@ -365,8 +364,6 @@ Additionally, this specification defines the following keys, which MAY be includ
 |`attribution`|An attribution to be displayed when the map is shown to a user. Implementations MAY decide to treat this as HTML or literal text. |string|
 |`type`|The type of the tileset |a string with a value of either `overlay` or `baselayer`|
 |`version`|The version number of the tileset|a string containing a valid version according to [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) |
-
-The JSON object MAY also include any other keys with an arbitrary value. This specification recommends nesting all application-specific data in an object under a semi-unique key to avoid overlap with other application-specific data or keys that may be defined in future versions of this specification. For example, instead of including the custom fields `author` and `companyId` directly in the top level of the metadata object, they SHOULD be nested in another object under a key with your project or organization name.
 
 ---
 
--- a/spec/v3/spec.md
+++ b/spec/v3/spec.md
@@ -21,26 +21,26 @@ An archive consists of five main sections:
 1. Optional leaf directories (described in [Chapter 4](#4-directories))
 1. The actual tile data
 
-These sections are normally in the same order as in the list above, but theoretically, it is possible to relocate all sections other than the header arbitrarily.
-The only two restrictions are that the header is at the start of the archive and the root directory MUST be contained in the first 16,384 bytes (16 KB) of the archive so that latency-optimized clients can retrieve the root directory in advance and ensure that it is complete.
+These sections are normally in the same order as in the list above, but it is possible to relocate all sections other than the header arbitrarily.
+The only two restrictions are that the header is at the start of the archive, and the root directory MUST be contained in the first 16,384 bytes (16 KiB) so that latency-optimized clients can retrieve the root directory in advance and ensure that it is complete.
 
 ```
-           Root Directory   Meta Data   Leaf Directories   Tile Data
+           Root Directory   Metadata    Leaf Directories   Tile Data
                Length        Length          Length         Length
-          <--------------> <---------> <----------------> <--------->
-+--------+----------------+-----------+------------------+-----------+
-|        |                |           |                  |           |
-| Header | Root Directory | Meta Data | Leaf Directories | Tile Data |
-|        |                |           |                  |           |
-+--------+----------------+-----------+------------------+-----------+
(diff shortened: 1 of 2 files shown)


bdon:
> PMTiles is a single-file archive format for tiled data. It enables low-cost, zero-maintenance map applications for "serverless" environments without having to rely on a custom tile backend or a third-party provider. This is achieved by packing all tiles of a tileset into an archive so that all tiles can be accessed easily and without much overhead via HTTP range requests. By combining all the tiles into one archive, hosting costs are kept low, as it is usually a lot cheaper to update one large file than to update thousands or even millions of small files.

This information would better go into the README of the project and not the format spec

> The JSON object MAY also include any other keys with an arbitrary value. This specification recommends nesting all application-specific data in an object under a semi-unique key to avoid overlap with other application-specific data or keys that may be defined in future versions of this specification. For example, instead of including the custom fields `author` and `companyId` directly in the top level of the metadata object, they SHOULD be nested in another object under a key with your project or organization name.

I don't think this is necessary, can we give some real-world use cases here?

in https://github.com/onthegomap/planetiler we already prefix keys with `planetiler:` instead of nesting them in an object
A valid consideration if if we eventually want to store CRS data like OGC TileMatrixSet, which could exist under a `crs` keyed object. If v4 or later of this spec conflicts with a user's own data, that situation is already addressed by the major version bump from v3 to v4

spec/v3/spec.md
```diff
@@ -6,41 +6,40 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 Please refer to the [change log](./CHANGELOG.md) for a documentation of changes to this specification.
 
+## 1 Abstract
 
-## 1 Abstact
-
-PMTiles is a single-file archive format for tiled data. It enables low-cost, zero-maintenance map applications for "serverless" environments without having to rely on a custom tile backend or a third-party provider. This is achieved by packing all tiles of a tileset into an archive so that all tiles can be accessed easily and without much overhead via HTTP range requests. By combining all the tiles into one archive, hosting costs are kept low, as it is usually a lot cheaper to update one large file than to update thousands or even millions of small files.
+PMTiles is a single-file archive format for tiled data.
```
DerZade, line 11: Most specs have a short abstract, introduction or motivation paragraph at the beginning. It ensures the spec works as a standalone document and not only in this repository. I also don't think there are any real disadvantages to keeping this, since it's less than 100 words.

spec/v3/spec.md
```diff
@@ -365,18 +365,16 @@ Additionally, this specification defines the following keys, which MAY be includ
 |`type`|The type of the tileset |a string with a value of either `overlay` or `baselayer`|
 |`version`|The version number of the tileset|a string containing a valid version according to [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) |
 
-The JSON object MAY also include any other keys with an arbitrary value. This specification recommends nesting all application-specific data in an object under a semi-unique key to avoid overlap with other application-specific data or keys that may be defined in future versions of this specification. For example, instead of including the custom fields `author` and `companyId` directly in the top level of the metadata object, they SHOULD be nested in another object under a key with your project or organization name.
-
```
DerZade, line 369: I just added this as an "insurance policy" to reduce the probability of breaking someone's code when adding a new property. So for example it could cause complications if we want to add `vector_layers` in a future version of this spec, but someone else already uses `vector_layers` with a different format.
So this is just a friendly reminder to not use common names, which is already best practice, but you know how lazy software engineers are 😛 

Prefixed values like you did in `planetiler` is also okay. It solves the same "problem". We can also change this paragraph to include something along the lines of _"Please prefix your keys in the root object"_. Would serve the exact same purpose of what I tried to do :)

spec/v3/spec.md
```diff
@@ -178,21 +178,21 @@ The field can have one of the following values:
...
+| `0x01` | MVT Vector Tile    |
 | `0x02` | PNG                |
 | `0x03` | JPEG               |
 | `0x04` | WebP               |
 | `0x05` | AVIF               |
 
 #### Min Zoom (MinZ)
 
-The Min Zoom is a 1-byte field specifying the minimum zoom (LOD) of the tiles.
+The Min Zoom is a 1-byte field specifying the minimum zoom of the tiles, >= 0.
```
DerZade, line 189: The `>= 0` is a bit redundant, since it is an unsigned integer.

spec/v3/spec.md
```diff
@@ -178,21 +178,21 @@ The field can have one of the following values:
...
 
 This field is encoded as an 8-bit unsigned integer.
 
 #### Max Zoom (MaxZ)
 
-The Max Zoom is a 1-byte field specifying the maximum zoom (LOD) of the tiles.
+The Max Zoom is a 1-byte field specifying the maximum zoom (LOD) of the tiles >= MinZ.
```
DerZade, line 195:
```suggestion
The Max Zoom is a 1-byte field specifying the maximum zoom (LOD) of the tiles. It must be greater than or equal to the min zoom.
```

bdon:
A long overdue update: I have some spec clarifications I need to make based on experience implementing `extract` in https://github.com/protomaps/go-pmtiles/issues/68

spec/v3/spec.md
```diff
@@ -335,25 +335,25 @@ Offsets are encoded either as `Offset + 1` or `0`, if they are equal to the sum
...
 
 _¹ Please refer to [Section 4.2](#42-encoding) for details on how Tile ID and Offset are encoded._
 
 ## 5 JSON Metadata
 
-The meta data section MUST contain a valid JSON object encoded in UTF-8, which MAY include additional meta data related to the tileset that is not already covered in the header section.
+The metadata section MUST contain a valid JSON object encoded in UTF-8, which MAY include additional metadata related to the tileset that is not already covered in the header section.
 
-If the [Tile Type](#tile-type-tt) in the header has a value of _Mapbox Vector Tile_, the object SHOULD contain a key of `vector_layers` as described in the [TileJSON 3.0 specification](https://github.com/mapbox/tilejson-spec/blob/22f5f91e643e8980ef2656674bef84c2869fbe76/3.0.0/README.md#33-vector_layers).
+If the [Tile Type](#tile-type-tt) in the header has a value of _MVT Vector Tile_, the object SHOULD contain a key of `vector_layers` as described in the [TileJSON 3.0 specification](https://github.com/mapbox/tilejson-spec/blob/22f5f91e643e8980ef2656674bef84c2869fbe76/3.0.0/README.md#33-vector_layers).
```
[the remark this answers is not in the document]
bdon: Mapbox is a trademark, so it needs to stay as-is.

spec/v3/spec.md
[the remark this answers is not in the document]
bdon: Thanks!

spec/v3/spec.md
```diff
@@ -123,15 +123,15 @@ This field is encoded as a little-endian 64-bit unsigned integer.
 
 #### Number of Addressed Tiles
 
-The Number of Addressed Tiles is an 8-byte field specifying the total number of tiles that are addressable in the PMTiles archive (before Run-Length encoding).
+The Number of Addressed Tiles is an 8-byte field specifying the total number of tile entries in the PMTiles archive, before RunLength Encoding.
```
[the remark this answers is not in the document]
bdon: Fixed

spec/v3/spec.md
bdon: I'm going to move that motivation paragraph to the README, some of the language is subjective - accessed easily, costs low, etc and doesn't fit in the spec.

PR #243, opened 2023-08-27
feat: add maplibre-gl-legend plugin

JinIgarashi:
I thought it would be useful for users to inspect the PMTiles dataset if maplibre legend is visible for each layer.

I have a maplibre-gl-legend plugin to simply add style legend on maplibre map.

https://github.com/watergis/maplibre-gl-legend

I just added the plugin to PMTiles viewer and it works quite well.

Viewer with this legend plugin looks like the below screenshots.

image

If `Only rendered` is checked, only show current rendered layers.

image

Note. This legend plugin lists all layers in maplibre style.json. The layers on top-right of this viewer is `source-layer`. The definition of layer between top-right panel and this plugin are a bit different.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -38,6 +38,7 @@
         "@types/react-dom": "^18.0.0",
         "@types/react-svg-pan-zoom": "^3.3.5",
         "@vitejs/plugin-react": "^1.3.0",
+        "@watergis/maplibre-gl-legend": "^2.0.4",
         "prettier": "^2.8.4",
         "typescript": "^4.6.3",
         "vite": "^2.9.16"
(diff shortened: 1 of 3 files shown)


bdon:
Thanks! this looks useful, however:

* we now have repeat information in two places between this and the layer toggle
* the legend shows `_point`, `_stroke` and `_fill` for every layer. This usually is not useful to the viewer of the map, it is an implementation detail of how all tile can be shown in MapLibre. 

What does seem useful is having the color from the legend. Can that be implemented on the right side toggle menu? 

Are there some features I'm missing that the legend plugin adds?

JinIgarashi:
I agree your points. I am also concerned that there are two duplicated information.

I could do add legend to existing top right panel. But it’s difficult for me since I don’t have a good knowledge of react.

If `showCheckbox` is set as true, the plugin also can have a feature to switch layer visibility.
As default, it shows exactly layer ID in maplibre, but we can also give other names instead of showing layer ID by using first argument of plugin constructor.

another option might be to use this layer switching feature of plugin and remove existing layer panel from top right of viewer.

PR #249, opened 2023-09-04
Correct TileID example

bchapuis:
This example does not correspond to the test cases provided in the javascript reference implementation.

https://github.com/protomaps/PMTiles/blob/17583ae19dba1392f48e63a12669f834064e797e/js/test/v3.test.ts#L40


bdon:
Thanks for the catch!

PR #263, opened 2023-10-02
Spec 3.3

bdon:
Making an executive decision here and adding some constraints to the existing spec based on discoveries in `pmtiles extract`:

* `clustered` is only an absolute ordering if the first tile is at offset 0. All the existing implementations I know of do this already.
* Multiple levels of leaf directories have always been implicitly supported, because there wasn't any extra logic needed. However, multiple leaf directories breaks the absolute ordering of the data in the leaf section (breadth-first or depth-first?), and the situations needed - very large archives with billions+ of unique tiles - I have not encountered in a year of V3 being out.

cc @DerZade

### DIFF
--- a/spec/v3/CHANGELOG.md
+++ b/spec/v3/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Version 3.3
+- Narrow `clustered` meaning: first tile entry must be offset 0.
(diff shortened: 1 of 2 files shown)


bdon:
Added recommended MIME type to spec.

Issue #247, opened 2023-09-04
Maplibre adapter should pass along attribution


jleedev:
https://github.com/protomaps/PMTiles/blob/17583ae19dba1392f48e63a12669f834064e797e/js/adapters.ts#L129-L134

Should also include: `attribution: m.attribution,`, where `m` is the result of `instance.getMetadata()`. Or the TileJSON could be the entire metadata verbatim, plus the fields from the header.

Maplibre would then be able to show the attribution automatically.

bdon:
I'm a bit torn on this because the size of metadata is unbounded. This is not a theoretical concern - tippecanoe automatically produces a `tilestats` key with info on column values, which can easily be 30 kb compressed - making the map load block on that to load a single sentence, by default, seems bad.

jleedev:
Yes, that's unfortunate.

For some numbers:

- The planet at tile.ourmap.us contains (approximately) 39,359 bytes of json, or 2,893 gzipped, which I expect should be typical of any users of the OpenMapTiles schema.
- The pmtiles from alltheplaces.xyz contains 37,636 bytes of _gzipped_ json, which decompresses to 141,885 bytes. (Since this is produced by throwing a huge amount of nonsense at tippecanoe without care for how big the tilejson becomes.)

On the one hand, fetching the first 64 KiB should therefore allow reading the complete metadata in many situations. On the other hand, the current spec and implementation only fetch the first 16 KiB in order to read the header and root. So this could be achieved by increasing the length of this initial read.

(The metadata could also be treated as a potentially-truncated JSON document to find the attribution, but that seems foolhardy.)

jleedev:
Manipulating the attribution of the maplibre source after it has loaded is … sort of possible but feels hacky. If maplibre had native pmtiles support it would be more plausible for the source to quickly read the root to determine the extents and format, and then lazily fetch the metadata.

bdon:
@jleedev the 16kb was intentionally chosen to be conservative - we can expand this to 32kb or 64kb if we discover that's a better default, while remaining backwards compatible with old archives for "free". The reverse is not true (if we started with a 32kb fetch, and later changed it to be 16kb, we'd have to add logic and extra requests to deal with old archives)

[...]

PR #266, opened 2023-10-16
PoC for loading metadata in maplibre adapter [#247]

bdon:


bdon:
Closing this because I don't feel that introducing options into the protocol string is preferable in the long term.

PR #270, opened 2023-10-18
Inspector: Explose true/false values

tordans:
`true`/`false` where hidden from the UI, they need to be transformed to strings.

Bildschirmfoto 2023-10-18 um 15 56 18


bdon:
@tordans this looks good! can you fix the CI failure related to typescript? `typeof` may return a string?

bdon:
@tordans thanks for the PR!

PR #273, opened 2023-10-23
Flip latitude and longitude in the spec.md

bramp:
Both the C++ and Python implementations decode this as bytes 0-3 as longitude and 4-7 as latitude. I assume the spec has a typo.

I discovered this while writing a dart implementation.


bramp:
@bdon you have updated the spec a couple of times recently. Mind looking at this? Thanks

bdon:
@bramp thanks!

PR #278, opened 2023-10-26
[ol-pmtiles] mvtOptions property to pass options to MVT Format

cieloazul310:
I propose to add a new property called `mvtOptions` to the options of `PMTilesVectorSource` in order to pass options to `MVT` format.

### DIFF
--- a/openlayers/src/index.js
+++ b/openlayers/src/index.js
@@ -78,13 +78,13 @@ export class PMTilesVectorSource extends VectorTile {
     });
   };
 
-  constructor(options) {
+  constructor({ mvtOptions, ...options }) {
     super({
       ...options,
       ...{
         state: "loading",
         url: "pmtiles://" + options.url + "/{z}/{x}/{y}",
-        format: new MVT(),
+        format: new MVT(mvtOptions),
       },
     });
 
--- a/openlayers/src/script_includes.js
+++ b/openlayers/src/script_includes.js
@@ -84,13 +84,13 @@ export class PMTilesVectorSource extends ol.source.VectorTile {
     });
   };
 
-  constructor(options) {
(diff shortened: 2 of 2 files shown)


bdon:
@cieloazul310 thanks, in what specific use cases do we need to change the defaults of `MVT`?

cieloazul310:
@bdon `MVT` format options allow filtering of the layers included in a VectorTile.

```ts
const layer = new VectorTileLayer({
  source: new VectorTileSource({
    format: new MVTFormat({
      layers: [], // layers to read
    }),
  }),
});
```


In an actual example, PMTiles, which the Geospatial Information Authority of Japan has started providing, contains 24 layers.

bdon:
@cieloazul310 thanks for the example - what is the advantage of filtering it via passing layers to `mvt` instead of simply not styling those layers?

cieloazul310:
@bdon Sure, it's possible to replace it with filtering within the style function. I tried both and saw no difference in performance. Thank you for responding to this matter!

bdon:
@cieloazul310 if this change does not enable any new functionality or optimization, then I would prefer to leave the code as-is for simplicity. Adding another options parameter is something we would have to maintain going forward. 

Is there another aspect I'm missing?

cieloazul310:
No problem. I noticed that by not adding the options parameter, the current `PMTilesVectorSource` could simply extend `VectorTileSource` to maintain type compatibility.

prusswan:
In case anyone is looking for the actual pmtiles, it is at: https://cyberjapandata.gsi.go.jp/xyz/optimal_bvmap-v1/optimal_bvmap-v1.pmtiles

https://github.com/optgeo/c2/blob/main/Rakefile

PR #304, opened 2023-11-22
Support for Last-Modified header

tengl:
This PR adds support for the Last-Modified header in PMTiles. Each serverless implementation needs to add support for it as well. 

Use cases:
Azure CDN standard [currently does not support ETag](https://learn.microsoft.com/en-us/azure/cdn/cdn-how-caching-works#validators). (See  #197 for Azure support)

### DIFF
--- a/js/index.ts
+++ b/js/index.ts
@@ -214,6 +214,7 @@ export interface Header {
   centerLon: number;
   centerLat: number;
   etag?: string;
+  lastModified?: string;
 }
 
 export function findTile(entries: Entry[], tileId: number): Entry | null {
@@ -246,6 +247,7 @@ export function findTile(entries: Entry[], tileId: number): Entry | null {
 export interface RangeResponse {
   data: ArrayBuffer;
   etag?: string;
+  lastModified?: string;
   expires?: string;
   cacheControl?: string;
 }
@@ -356,6 +358,7 @@ export class FetchSource implements Source {
     return {
       data: a,
       etag: resp.headers.get("ETag") || undefined,
+      lastModified: resp.headers.get("Last-Modified") || undefined,
       cacheControl: resp.headers.get("Cache-Control") || undefined,
       expires: resp.headers.get("Expires") || undefined,
     };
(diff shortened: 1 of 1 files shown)


tengl:
@bdon I created a new PR because it found out that Azure CDN does not support `ETag` yet. I need it to optimize tile delivery by CDN with #197.

bdon:
Thanks for the PR. Azure Blob Storage does support `Etag`, right? So this only affects the interaction between the Serverless Function and any CDN in front. The purpose of `ETag` in the code is to ensure that any locally cached parts of the file can be invalidated when the data for the key changes in-place.

What is the practical effect of returning `Last-Modified` to the CDN? if the Azure CDN gets the same `Last-Modified` in the response for a given tile it could treat that as a no-op, but because it is transferring all the tile bytes anyways that seems like an unnecessary optimization. Is there something I'm missing?

tengl:
Yes, it supports ETag, both when used with `FetchSource` and `AzureStorageSource`. I saw that it was used in the cache, I don't think that will be affected. 

The code in #197 will check `Last-Modified` and compare it to the request `If-Modified-Since`, if equal it will return a `304`. Since it is possible to do this on the header, no extra bytes will be transferred. 

I have not tested this all the way through Azure CDN, just that the Azure Function return `304` when it should (but what else can the CDN return?).

tengl:
Aha, I tested the CDN now, and it returns `304` if I add `If-Non-Match` with a valid ETag and it hits the cache. So I guess Azure CDN does support ETags after all. 

But I guess this could still be useful, right? It is up to the implementation to decide how to use it.

bdon:
I would need a concrete example of where it is useful - the ETag is explicitly used as a cache key right now for correctness. Otherwise we need to maintain this extra data in the decoder state indefinitely into the future. 

From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

> Less accurate than an [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) header, it is a fallback mechanism.

> Last-Modified is also used by [crawlers](https://developer.mozilla.org/en-US/docs/Glossary/Crawler) to adjust crawl frequency, by browsers in [heuristic caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#heuristic_caching), and by content management systems (CMS) to display the time the content was last modified.

The CMS case is interesting, but if you want to determine "data date" of a PMTiles archive then embedding a timestamp into the metadata section is much more reliable than your blob storage.

tengl:
Yes, you are right. I initially had an idea that it was up to the source to return a header, but the source doesn't know anything about the PMTiles header so that won't work. 

Anyway, ETag seems to be supported in Azure Cache, at least the version I'm using so my use case is supported without Last-Modified. I'll just close the PR.

PR #305, opened 2023-11-24
Uses wrangler deploy instead of deprecated publish

cboone:
Tiniest of tiny changes. Just got sick of the deprecation warning. Not sure if you want to bump the nonexistent version.

Also, hi @bdon! 👋


bdon:
Thanks @cboone! long time no see :P

PR #311, opened 2023-12-13
Fixes units in Python writer default bounds

daniel-j-h:
When there are no user-provided bounds provided in the writer (as in: min/max_lat/lng_e7 properties) the writer mistakenly defaults to -180,-90,180,90 for these properties.

https://github.com/protomaps/PMTiles/blob/e1228f5df1a4ac852e0117dc08c8813089a5c8af/python/pmtiles/tile.py#L277-L283

But these properties need to be multiplied by 10000000 as seen e.g. in this examples and as indicated by the "_e7" postfix

https://github.com/protomaps/PMTiles/blob/e1228f5df1a4ac852e0117dc08c8813089a5c8af/python/examples/create_raster_example.py#L29-L32

Without this fix we were seeing pmtiles with bounds around null island, because the order of magnitude was wrong.


bdon:
Good catch, thanks!

Issue #248, opened 2023-09-04
AstroJS build error "Named export 'Protocol' not found."


tordans:
For some reason, I imported `{ Protocol }` and not `* as pmtiles` (like this https://github.com/tordans/berlin-bikenetwork-monitoring/blob/0704fe8a517e75669154e1bf2a9e5ec0ca810da8/src/components/Map/Map.tsx#L3) in https://github.com/tordans/berlin-bikenetwork-monitoring (maybe I copied it from [this one place](https://github.com/search?q=repo%3Aprotomaps%2FPMTiles+%22new+Protocol%22&type=code)? IDK).

When I build this AstroJS project via `npm run build` I get the following error.

```
error   Named export 'Protocol' not found. The requested module 'pmtiles' is a CommonJS module, which may not support all module.exports as named exports.
  CommonJS modules can always be imported via the default export, for example using:

  import pkg from 'pmtiles';
  const { Protocol } = pkg;

file:///Users/…/berlin-bikenetwork-monitoring/dist/chunks/pages/index.astro.fa247295.mjs:6
import { Protocol } from 'pmtiles';
         ^^^^^^^^
SyntaxError: Named export 'Protocol' not found. The requested module 'pmtiles' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'pmtiles';
const { Protocol } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
```

---

When I do the suggested change, I get…

```
"default" is not exported by "node_modules/pmtiles/dist/index.mjs", imported by "src/components/Map/Map.tsx".
file: /Users/…/berlin-bikenetwork-monitoring/src/components/Map/Map.tsx:3:7
1: import maplibregl from 'maplibre-gl'
2: import 'maplibre-gl/dist/maplibre-gl.css'
3: import pmtiles from 'pmtiles'
          ^
4: import { useEffect } from 'react'
5: import ReactMapGl, { NavigationControl } from 'react-map-gl/dist/es5/exports-maplibre.js'
```

----

The solution is, to do [as the docs say](https://github.com/protomaps/PMTiles/blob/main/js/README.md?plain=1#L12) and use ` import * as pmtiles from "pmtiles";`.

However, using this way of importing always feels like an error. And there should be a way to allow for the other syntax as well, right?

[...]

PR #329, opened 2024-01-27
Fix package.json TS issue …

tordans:
… by blindly following what https://publint.dev/pmtiles@2.11.0 recomments which does solve https://github.com/protomaps/PMTiles/issues/248#issuecomment-1913028309

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -2,8 +2,9 @@
   "name": "pmtiles",
   "version": "2.11.0",
   "description": "PMTiles archive decoder for browsers",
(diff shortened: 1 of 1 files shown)


tordans:
Nope, not a fix, see https://github.com/protomaps/PMTiles/issues/248#issuecomment-1913031957

bdon:
I believe the preference nowadays is for ESM modules over CommonJS ones, but a CommonJS one is at `dist/index.cjs`

Issue #317, opened 2024-01-04
pmtiles in SSR in sveltekit v2, vite 5, Astro (ES6 module problems)


JinIgarashi:
I upgraded sveltekit to v2, then faced an issue of importing pmtiles in SSR in sveltekit v2 and vite5. The following error log was generated by vite when I built source code.

```
3:24:16 PM [vite] Error when evaluating SSR module /src/routes/+layout.server.ts: failed to import "/src/lib/server/helpers/index.ts"
|- SyntaxError: [vite] Named export 'PMTiles' not found. The requested module 'pmtiles' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'pmtiles';
const {PMTiles} = pkg;
```

But changing to the example suggested by the error does not solve the issue. I was really struggling to fix it, but somehow I managed to fix the issue to add `pmtiles` into `ssr.noExternal` at vite.config.ts as following example.

```ts
import { defineConfig } from 'vite'

export default defineConfig({
  plugins: [sveltekit()],
  ssr: {
    noExternal: [ 'pmtiles', 'maplibre-gl']
  }
})
```

after adding the above setting, works without changing. maplibre-gl also has same issue of importing at SSR, it also needs to be added to ssr.noExternal config.

I am not sure how many users use pmtiles in sveltekit, I wanted to share my solution with the community for future reference. This is a tricky issue.

bdon:
Thanks for the report, let's leave this open to help others with the issue and work on making a reproducible example with sveltekit2+vite5

Havegum:
Here's a minimal sveltekit project that demonstrates the strange import behaviour:
https://github.com/Havegum/pmtiles-sveltekit-bug

It seems to load differently in dev mode compared to at build time.

In dev mode, you can import the exported classes like this:
```js
import * as pmtiles from "pmtiles";
const { PMTiles } = pmtiles;
```

At build time, I've gotten a hold of the exported classes like so:
```js
import * as pmtiles from "pmtiles";
const { PMTiles } = pmtiles.default;
```

The build time syntax seems to imply that you could just import default from `pmtiles` directly, but then it complains in dev mode that `pmtiles` _has_ no default export 🤷 

### Workaround

For now, I get around this with the nullish coalescing operator:
```js
import * as pmtiles from "pmtiles";
const { PMTiles } = pmtiles.default ?? pmtiles;
```

bdon:
[...]

PR #330, opened 2024-01-27
Make package.json main es6 module instead of cjs [#317, #248]

bdon:
@tordans @Havegum @JinIgarashi I think this should resolve all the cases where Svelte/Astro find the CJS module before the ESM one. However, what's left to be tested is if this still works correctly for the cases needing CJS (I think NodeJS 16 at least) so we need to test those more. 

Any insight into the best choices in package.json or references to other libraries is appreciated

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -16,7 +16,24 @@
         "esbuild": "^0.15.11",
         "esbuild-runner": "^2.2.1",
         "prettier": "^2.8.4",
+        "tsx": "^4.7.0",
         "typescript": "^4.5.5"
+      }
+    },
+    "node_modules/@esbuild/aix-ppc64": {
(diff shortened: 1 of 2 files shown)


JinIgarashi:
@bdon Thank you for fixing this. New package.json looks good. 
i want to test this version of PMTiles package in our GeoHub repo without additional ssr.noExternal settings in vite config. 
is it possible to release this branch as a beta version before bumping version officially? So I can test it.

About Node version, 16 maybe does not need to be supported anymore. Now 20 is LTS, and 18 is maintenance. Support for 16 already ended last year. See the below page

https://nodejs.org/en/about/previous-releases

bdon:
blocked on https://github.com/folke/esbuild-runner/issues/36

Issue #287, opened 2023-11-09
Open to JS code improvements/restructuring


sachaw:
Hi, I use this library quite frequently, would you be open to receiving a PR that includes a bunch of project related improvements;

Use PNPM, modern exports, auto linting etc?

I'll prep the changes and if there there are specifics you don't agree with, I can revert them.

Thanks.

bdon:
1. not if it requires all users to use `pnpm`
2. for modern exports, do you mean changing the exports? this would be a breaking semver change
3. I have it on my todo list to add a lint check, do you prefer eslint?
Thanks for asking.

sachaw:
1. Users would not have to use pnpm, it would be preferred for development only.
2. Modern exports are not a breaking change, the old usage is still maintained
3. I have a preference for Biome as the longer, requires only a single config file and not the half dozen eslint plugins to get it working

Additionally, making this a native ESM library would be preferable and if people insist on using cjs they can just transpile this dependency.

bdon:
> Additionally, making this a native ESM library would be preferable and if people insist on using cjs they can just transpile this dependency.

Isn't this already the case? The cjs etc targets are created by esbuild here:

https://github.com/protomaps/PMTiles/blob/main/js/package.json#L14

sachaw:
There's a few more things required that's outlined well here: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

Also making it an ESM only package is desirable now to encourage people to stop using cjs as it's a constant burden on the ecosystem. Instead of us transpiring it, let them do it, ESM has support as far back as node 12, we should only be supporting 18, and soon 20 (current LTS).

sachaw:
WIP repo: https://github.com/sachaw/PMTiles

bdon:
Thanks for the example. We need to stay on Node 18 instead of 20 because AWS Lambda doesn't have Node 20 yet.

In general, I'm OK with the packaging set-up as is although it may not align with every individual developer's preferences. Ideally it should use as few devDependencies as possible to minimize the possibility of situations across all the platforms (all browsers + node + lambda + cloudflare workers + demo).

Is there a specific pain point you intend to resolve with your changes that we might find another solution for?

sachaw:
I am of the same mindset, minimal dependencies and minimal maintenance burden (reduced dev environment complexity). node 18 is fine, just keeping up with LTS is the important part, not necessarily running the current release.

[...]

PR #343, opened 2024-02-01
add typedoc [#287]

bdon:


js/index.ts
```diff
@@ -834,6 +837,8 @@ export class SharedPromiseCache {
   }
 }
 
+// if source is a string, create a browser FetchSource().
```
HarelM, line 840: This should be using /** type of comments to show in the docs as far as I know.

HarelM: Might be worth adding this on the conductor method too, this way you would see it in vscode when you write the code to use the constructor.

HarelM:
BTW, what are you doing with the generated docs? Are you publishing then to github pages?
If not, you might not need anything soecial as when you create the .d.ts file the comments with be there, and the IDE will pick it up.
So typedoc might be an overkill here.
IDK... ping me on slack if you want to pick my brain.

bdon:
Not 100% sure if types are even working right now - can you confirm with 3.0.0-alpha.2? https://publint.dev/pmtiles@3.0.0-alpha.2

I plan to add the docs to GitHub pages, seems helpful for google-ability as well as for those not using IDEs

HarelM:
I don't see any comments here:
https://unpkg.com/browse/pmtiles@3.0.0-alpha.2/dist/index.d.ts
Also there's an empty index2.d.ts file, see here:
https://unpkg.com/browse/pmtiles@3.0.0-alpha.2/dist/

bdon:
thanks, removed the index2 file. I haven't added any comments yet, hence the typedoc isn't published

HarelM:
The d.ts file should contain comments straight from the typescript code, regardless of typedoc.

HarelM:
But it requires "/**" type of comments, I think.

PR #357, opened 2024-02-08
fix reader.py reading all the tiles

gsimko:
currently it doesn't yield tiles in a run sequence, only the first one

### DIFF
--- a/python/pmtiles/reader.py
+++ b/python/pmtiles/reader.py
@@ -63,9 +63,10 @@ def traverse(get_bytes, header, dir_offset, dir_length):
     entries = deserialize_directory(get_bytes(dir_offset, dir_length))
     for entry in entries:
         if entry.run_length > 0:
-            yield tileid_to_zxy(entry.tile_id), get_bytes(
-                header["tile_data_offset"] + entry.offset, entry.length
-            )
+            for i in range(entry.run_length):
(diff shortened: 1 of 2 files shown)


gsimko:
Note that I don't know the larger scope how reader is used.
pmtiles-convert is broken without this because it will miss many tiles on the output

bdon:
Thanks, can you add a unit test that covers this to https://github.com/protomaps/PMTiles/blob/main/python/test/test_reader_writer.py ?

> Note that I don't know the larger scope how reader is used.

I don't know either, currently this python implementation is superseded by https://github.com/protomaps/go-pmtiles which you can use for converting MB -> PMTiles, but doesn't have 100% of the convert functionality

pkage:
Just ran into this issue today as well—thank you for the fix @gsimko!

This also breaks the `pmtiles` python package. We're using it internally to store georeferenced ML training data and this has caused some subtle bugs, and would be interested in getting this fix merged ASAP.

Happy to take the lead on writing the test @gsimko, just let me know.

gsimko:
I've added a unit test and verified that it was broken before the fix.

I'd use go-pmtiles but it only works in the pmtiles->mbtiles direction and sometimes I need to do the opposite.

bdon:
Thanks for the fix!

bdon:
3.2.1 uploaded: https://pypi.org/project/pmtiles/

PR #364, opened 2024-02-20
Remove print statements in python code

JesseCrocker:
Print statements are unexpected in a library context, so they were removed. The other option would be to convert them to Logging calls, but since there is no other use of Logging, that seems like part of a larger project.


bdon:
Thanks!

Issue #49, opened 2022-06-14
Inspector app improvements


bdon:
- [x] SVG tile previews should be zoomable
- [x] Should be able to drill down into leaf directories
- [x] should be able to preview vector tiles in leaflet
- [x] SVG should be feature-level inspectable
- [x] map preview should be feature-level inspectable
- [x] see directory Len and header-level metadata
- [x] inspect SVG with mismatched extents
- [x] correctly read tile and map hash states
- [x] split out "individual tile inspector" to separate page
- [x] view raw zxy and TIleJSON endpoints

bdon:
See if we can reuse https://github.com/acalcutt/maplibre-gl-inspect

bdon:
* Display labels from TileJSON

bdon:
I'd like to port pmtiles.io to use https://github.com/maplibre/maplibre-gl-inspect , since that's what maps.protomaps.com and Maputnik are using.

wipfli:
Other nice feature would be to load a pmtiles url not only on click on load button, but also on press ENTER in the url text input field

bdon:
The inspector should also accept TileJSON and full Style JSONs to display, just for convenience. Raw Z/X/Y URLs won't work since we don't have enough metadata to derive `vector_layers` for inspection.

prusswan:
> I'd like to port pmtiles.io to use https://github.com/maplibre/maplibre-gl-inspect , since that's what maps.protomaps.com and Maputnik are using.

Actually, I think both implementations have their merits (I like how pmtiles.io works with popup freezing and treats each source layer as a group, with possibly different geometry types). For the metadata, maplibre-gl-inspect only needs a list of layers present in the source (e.g. ["water",...}]).

btw, I noticed that switching from Map View to Tile Inspector does not work for local files (you have to drag the file again)

bdon:
Yeah, the plan is to keep tile inspector as a separate page instead of all under one single page app, because trying to manage the hash state gets too complicated.

bdon:
It would be helpful to perform light validation of the PMTiles header/metadata to assist implementers

bdon:
Make sure #537 is supported

bdon:
Closing this old mega-issue in favor of a feedback issue on the new http://beta.pmtiles.io

PR #379, opened 2024-03-15
viewer: break tile inspect into separate entrypoint [#49]

bdon:


JesseCrocker:
This seems to have broken something. When i click the "Tile Inspector" button on https://protomaps.github.io/PMTiles/ it takes me to https://protomaps.github.io/PMTiles/tileinspect/?url=my-file.pmtiles, which returns a 404 page from github.

It seems like splitting the tile inspector into a separate entrypoint is also problematic for inspecting local files, because they need to be opened by the user, not just have a url passed.

bdon:
@JesseCrocker it should be fixed now at https://protomaps.github.io/PMTiles/tileinspect/

It will work for local files because it re-uses the same selector start page as the map view. By splitting it into a different URL we can also add the Z/X/Y to the URL state, which I haven't done yet.

JesseCrocker:
The 404 is gone, and with remote urls it works properly, but when I use a local file, then click "tile inspector" i just get the header, but no tile info and no way to choose a file.

Screenshot 2024-03-25 at 3 18 22 PM

bdon:
We should make the inspector link just link to the landing page for local files, since there isn't any way to transfer the permissions across a page load.

PR #383, opened 2024-04-10
Update js/examples/maplibre.html

paulovieira:
Some small changes to make the this important example more clear:
- wrap all the code in an async function, so that we can use `await p.getHeader()`
- some small changes for coherence (double quotes, etc)

If there is interest I can make these same changes in the other 2 examples (maplibre_raster_dem.html and leaflet.html)

### DIFF
--- a/js/examples/maplibre.html
+++ b/js/examples/maplibre.html
@@ -17,22 +17,23 @@
     <body>
         <div id="map"></div>
         <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


bdon:
I don't view using `await` as inherently more correct than using `.then()` in this context of a simple script using the IIFE script-includes. Is there another reason why this would be better?

bdon:
Fixed formatting in #438 , thanks.

Issue #407, opened 2024-06-11
Allow Customization of OpenLayers Format Parameters


vycius:
### Problem

The PMTiles OpenLayers library currently always sets the `format` to MVT with default parameters in the code found here: [link](https://github.com/protomaps/PMTiles/blob/main/openlayers/src/script_includes.js#L92). 

This hardcoding of the MVT format limits the ability to customize OpenLayers format parameters, such as setting `idProperty` or changing `featureClass`, which are essential for enabling full editing and geometry support.

### Proposed Solution

Modify the library to allow customization of the format and the passing of additional arguments. This would enable users to specify the MVT parameters as needed, providing greater flexibility and control over their OpenLayers configurations.

### Current Workaround

A current workaround involves setting the format after initializing the `PMTilesVectorSource`. For example:
```ts
(layer.getSource() as PMTilesVectorSource).format_ = new MVT({
  featureClass: Feature,
  idProperty: 'code',
});
```
While this method works, it is not an elegant solution. It would be preferable to be able to pass these parameters directly to the `PMTilesVectorSource` constructor.

bdon:
v0.5.0 published on NPM: https://www.npmjs.com/package/ol-pmtiles

Thanks!

bdon:
FYI, there's now a version 1.0 with TypeScript and ESM support: https://www.npmjs.com/package/ol-pmtiles

PR #408, opened 2024-06-11
Allow Customization of OpenLayers Format Parameters

ambrazasp:
As described in #407, `format` option is set as `new MVT()` for OpenLayers.

However in some cases you need to rewrite `format` and pass MVT options. 

There is two solutions:
- Pass MVT options
- Pass custom formater

In this case it's easier to go with 2nd option (passing custom formatter).

From now on you can do something like this:
```js
new PMTilesVectorSource({
  ..., 
  format: new MVT({
    featureClass: ol.Feature,
    idProperty: 'customId',
  }),
})
```

### DIFF
--- a/openlayers/src/index.js
+++ b/openlayers/src/index.js
@@ -78,11 +78,14 @@ export class PMTilesVectorSource extends VectorTile {
   };
 
   constructor(options) {
-    options.state = "loading"
-    options.url = "pmtiles://" + options.url + "/{z}/{x}/{y}"
-    options.format = options.format || new MVT();
-
-    super(options);
+    super({
+      ...options,
+      ...{
+        state: "loading",
+        url: "pmtiles://" + options.url + "/{z}/{x}/{y}",
+        format: options.format || new MVT(),
+      },
+    });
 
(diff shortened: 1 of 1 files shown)


openlayers/src/index.js
```diff
@@ -78,14 +78,11 @@ export class PMTilesVectorSource extends VectorTile {
...
-    super({
-      ...options,
-      ...{
-        state: "loading",
-        url: "pmtiles://" + options.url + "/{z}/{x}/{y}",
-        format: new MVT(),
-      },
-    });
+    options.state = "loading"
+    options.url = "pmtiles://" + options.url + "/{z}/{x}/{y}"
```
bdon, line 82: This doesn't seem to work because line 88, the first argument to `pmtiles.FetchSource` needs to be the plain URL `https://example.com/name.pmtiles` instead of the prefixed one that `PMTilesVectorSource` expects

Should we leave the code like before but just change l86 to: `format: options.format || new MVT(),` ?

Thanks

ambrazasp: Good point. Changed to old version. Only `format` has a fallback (as new MVT())

PR #409, opened 2024-06-11
Add CACHE_CONTROL env, deprecate CACHE_MAX_AGE 

MarkusWendorf:
Hey there

As [discussed](https://github.com/protomaps/PMTiles/discussions/404) I would like to deprecate the `CACHE_MAX_AGE` option in favour of `CACHE_CONTROL`. 

Users would then be able to set any `CACHE_CONTROL` directive they might need like `stale-while-revalidate` or `s-maxage`.

The change to add this feature would be rather small, however I had too much time on a train ride and improved a few things for readability. 

Changes:
- give variables meaningful names (avoid `o`, `p`, `a`, `e` etc.)
- restructure the file so one can read it from top to bottom (env first, globals, handler entrypoint, data source implementation)
- add types for environment variables in node.js
- remove type casts (type narrowing with `if`)
- use `function` keyword for all top level functions (instead of mixing `const fn = () => {}` and `function fn() {}`)

I know that my changes have gone a little far for this feature 😅, but I think this would also help future contributors (I'm open to split the PR in `feature` and `code readability`).

Thank you

### DIFF
--- a/serverless/aws/.gitignore
+++ b/serverless/aws/.gitignore
@@ -1,1 +1,2 @@
 dist
+node_modules
--- a/serverless/aws/src/index.ts
+++ b/serverless/aws/src/index.ts
@@ -58,11 +58,11 @@ export async function handler(
 // Assumes event is a API Gateway V2 or Lambda Function URL formatted dict
 // and returns API Gateway V2 / Lambda Function dict responses
 // Does not work with CloudFront events/Lambda@Edge; see README
-export const handlerRaw = async (
+export async function handlerRaw(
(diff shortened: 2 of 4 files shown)


bdon:
@MarkusWendorf can you split this into the feature for `CACHE_CONTROL` PR and then the refactoring after that? That will let me merge the feature and update the docs, etc sooner. Thanks

MarkusWendorf:
Closed in favor of https://github.com/protomaps/PMTiles/pull/423

PR #419, opened 2024-06-28
Update spec.md

CraigglesO:
the spec requires that `vector_layers` exists.

 - [x] Please do not open PRs for new features if you're a first time contributor; first open an issue to confirm that the new feature is appropriate.


bdon:
Good catch, thanks.

PR #423, opened 2024-07-08
Add CACHE_CONTROL env, deprecate CACHE_MAX_AGE

MarkusWendorf:
Hey there

As [discussed](https://github.com/protomaps/PMTiles/discussions/404) I would like to deprecate the CACHE_MAX_AGE option in favour of CACHE_CONTROL.

Users would then be able to set any CACHE_CONTROL directive they might need like stale-while-revalidate or s-maxage.

Supersedes https://github.com/protomaps/PMTiles/pull/409

### DIFF
--- a/serverless/aws/.gitignore
+++ b/serverless/aws/.gitignore
@@ -1,1 +1,2 @@
 dist
+node_modules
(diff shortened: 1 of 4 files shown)


bdon:
Docs updated: https://github.com/protomaps/docs/pull/45

Thanks!

Issue #338, opened 2023-12-14
CLI tool to convert raster to pmtiles


daniel-j-h:
The protomaps format allows for raster tile hierarchies.

We've been working with bundling up raster tiles recently and ended up using the python library from https://github.com/protomaps/PMTiles to do so (see also example in https://github.com/protomaps/PMTiles/blob/e1228f5df1a4ac852e0117dc08c8813089a5c8af/python/examples/create_raster_example.py).

What would have made that workflow simpler is a cli tool like `pmtiles convert` but instead of converting mbtiles into pmtiles it would take e.g. a slippy map tile hierarchy as in `z/x/y.ext` or a GeoTIF raster and turn it into a pmtiles file.

Should the convert sub-command be extended to support raster, too?

bdon:
issue transferred to main PMTiles repository

mabhub:
> What would have made that workflow simpler is a cli tool like `pmtiles convert` but instead of converting mbtiles into pmtiles it would take e.g. a slippy map tile hierarchy as in `z/x/y.ext` or a GeoTIF raster and turn it into a pmtiles file.

Isn't it already kinda simple by the use of [mb-util](https://github.com/mapbox/mbutil) which allow to convert tiles file tree to mbtiles, then converting it to pmtiles with `pmtiles convert`?

```shell
$ mb-util my-tiles-root-directory my_tiles.mbtiles
$ pmtiles convert my_tiles.mbtiles my_tiles.pmtiles
```

neodescis:
> > What would have made that workflow simpler is a cli tool like `pmtiles convert` but instead of converting mbtiles into pmtiles it would take e.g. a slippy map tile hierarchy as in `z/x/y.ext` or a GeoTIF raster and turn it into a pmtiles file.
> 
> Isn't it already kinda simple by the use of [mb-util](https://github.com/mapbox/mbutil) which allow to convert tiles file tree to mbtiles, then converting it to pmtiles with `pmtiles convert`?
> 
> ```shell
> $ mb-util my-tiles-root-directory my_tiles.mbtiles
> $ pmtiles convert my_tiles.mbtiles my_tiles.pmtiles
> ```

This is the route I went to convert some terrain tiles. It is indeed pretty simple, but:
1. This is not an obvious path to take for someone who doesn't know anything about mbtiles
2. It seems bad for pmtiles to depend on mb-util like this to get archives created

larsmaxfield:
[...]

PR #431, opened 2024-08-09
Add disk_to_pmtiles: Convert a directory of tiles to pmtiles

larsmaxfield:
- [x] Please do not open PRs for new features if you're a first time contributor; first open an issue to confirm that the new feature is appropriate.

See feature request at https://github.com/protomaps/PMTiles/issues/338.

### DIFF
--- a/python/pmtiles/convert.py
+++ b/python/pmtiles/convert.py
@@ -300,7 +300,9 @@ def disk_to_pmtiles(directory_path, output, maxzoom, **kwargs):
 
         # read tiles in ascending tile order
         count = 0
-        count_step = (2**(maxzoom-3))**2
+        if verbose:
+            count_step = (2**(maxzoom-3))**2 if maxzoom <= 9 else (2**(9-3))**2
+            print(" Begin writing %s to .pmtiles ..." % (n_tiles), flush=True)
         for tileid, filepath in tileid_path_set:
             f = open(filepath, 'rb')
             data = f.read()
@@ -309,8 +311,8 @@ def disk_to_pmtiles(directory_path, output, maxzoom, **kwargs):
                 data = gzip.compress(data)
             writer.write_tile(tileid, data)
             count = count + 1
-            if (count % count_step) == 0 and verbose:
-                print(" %s tiles inserted of %s" % (count, n_tiles))
+            if verbose and (count % count_step) == 0:
+                print(" %s tiles inserted of %s" % (count, n_tiles), flush=True)
 
         if verbose and (count % count_step) != 0:
             print(" %s tiles inserted of %s" % (count, n_tiles))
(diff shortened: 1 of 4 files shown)


larsmaxfield:
Working on fixing the metadata for disk_to_pmtiles.

larsmaxfield:
Fixed.

python/pmtiles/convert.py
```diff
@@ -178,3 +178,167 @@ def pmtiles_to_dir(input, output):
...
+    # Collect a set of all tile IDs
+    z_set = []  # List of all zoom levels for auto-detecting maxzoom.
+    tileid_path_set = []  # List of tile (id, filepath) pairs
+    zoom_dirs = get_dirs(directory_path)
+    zoom_dirs.sort(key=len)
+    try:
+        collect_max = int(maxzoom)
+    except ValueError:
+        collect_max = 99
+    collect_min = metadata.get("minzoom", 0)
```
bdon, line 229: What tool is creating metadata.json? https://github.com/felt/tippecanoe writes the minzoom and maxzoom as strings, so these may need to be `int(metadata.get...))`

example tippecanoe command: `./tippecanoe --output-to-directory foo ne_10m_admin_0_label_points.fgb --maximum-zoom 5 --force`

bdon: tippecanoe also writes a key called `json` that has `vector_layers`. I think this code will need to special-case `format=pbf` in order to translate `vector_layers` correctly as they are required in the spec:

https://github.com/protomaps/PMTiles/blob/main/spec/v3/spec.md#5-json-metadata

The other key it writes is `tilestats` - so all top-level keys of the `json` value could lifted up one level in the pmtiles metadata.

Tools other than tippecanoe might write different `metadata.json`?

bdon: If the above sounds like overkill then maybe we limit the disk_to_pmtiles to raster formats only (if that's the use case you need)

larsmaxfield: Ah I wasn't aware of that. Thanks for the explanation. 

For now let's limit it to raster formats. That's indeed the only use case I need. If someone with a need (and knowledge) for other formats wishes to add support, there's at least this function for them to add to.

larsmaxfield: Are you OK if I simply remove 'pbf` from the help and docstrings of the functions?

Or would you like an explicit rejection for formats not 'png', 'jpeg', 'webp', or 'avif'?

bdon: removing `pbf` is fine.

larsmaxfield: References to 'pbf' are now gone, and I added 'raster format' to be more specific.

bdon:
3.4.0 is published to https://pypi.org/project/pmtiles/ now, thanks!

larsmaxfield:
Nice and efficient! Happy to help.

PR #435, opened 2024-08-27
Update CloudFormation template.

bdon:
Updates to make the CF template as one-click as possible.

Not yet inlining the Lambda zip, that will come in a future commit.

Comments inline

@Honricris @charliemcgrady

### DIFF
--- a/serverless/aws/package.json
+++ b/serverless/aws/package.json
@@ -7,12 +7,13 @@
     "@types/node": "^18.11.2",
     "esbuild": "^0.20.0",
     "tsx": "^4.7.0",
-    "typescript": "^4.8.4"
+    "typescript": "^4.8.4",
+    "yaml-cfn": "^0.3.2"
   },
   "private": true,
   "scripts": {
     "tsc": "tsc --noEmit --watch",
-    "build": "esbuild src/index.ts --target=es2020 --outfile=dist/index.mjs --format=esm --bundle --platform=node --target=node18 --external:@aws-sdk/client-s3 --external:@aws-sdk/node-http-handler --banner:js=//$(git describe --always) && cd dist && zip lambda_function.zip index.mjs",
+    "build": "esbuild src/index.ts --target=es2020 --minify --outfile=dist/index.js --format=cjs --bundle --platform=node --target=node20 --external:@aws-sdk/client-s3 --external:@aws-sdk/node-http-handler --banner:js=//sha:$(git describe --always)",
     "test": "tsx ../shared/index.test.ts",
     "biome": "biome check --config-path=../../js/ src --apply",
     "biome-check": "biome check --config-path=../../js src"
--- a/serverless/aws/package.json
+++ b/serverless/aws/package.json
@@ -13,7 +13,8 @@
   "private": true,
   "scripts": {
     "tsc": "tsc --noEmit --watch",
-    "build": "esbuild src/index.ts --target=es2020 --minify --outfile=dist/index.js --format=cjs --bundle --platform=node --target=node20 --external:@aws-sdk/client-s3 --external:@aws-sdk/node-http-handler --banner:js=//sha:$(git describe --always)",
+    "build-zip": "esbuild src/index.ts --target=es2020 --outfile=dist/index.mjs --format=esm --bundle --platform=node --target=node20 --external:@aws-sdk/client-s3 --external:@aws-sdk/node-http-handler --banner:js=//$(git describe --always) && cd dist && zip lambda_function.zip index.mjs",
+    "build-cloudformation-stack": "esbuild src/index.ts --target=es2020 --minify --outfile=dist/index.js --format=cjs --bundle --platform=node --target=node20 --external:@aws-sdk/client-s3 --external:@aws-sdk/node-http-handler --banner:js=//sha:$(git describe --always) && tsx inline_lambda.ts",
     "test": "tsx ../shared/index.test.ts",
     "biome": "biome check --config-path=../../js/ src --apply",
     "biome-check": "biome check --config-path=../../js src"
--- a/serverless/aws/protomaps-template.yaml
+++ b/serverless/aws/protomaps-template.yaml
@@ -1,36 +1,39 @@
 AWSTemplateFormatVersion: '2010-09-09'
-Description: CloudFormation template to create a protomaps infraestructure to serve tiles.
+Description: Serve Z/X/Y tiles through CloudFront + Lambda from an existing S3 bucket.
 Parameters:
   BucketName:
-    Description: 'The name of the S3 bucket where you will store pmtiles files to be served (must be globally unique)'
+    Description: 'Name of an existing S3 bucket with .pmtiles tilesets. Should be in the same region as your CloudFormation stack.'
     Type: String
+    MinLength: 1
 
-  CodeBucketName:
-    Description: 'The S3 bucket name where the Lambda function code is stored (e.g., lambda-protomaps-code)'
-    Type: String
-
-  CodeKey:
-    Description: 'The S3 key for the Lambda function code (e.g., lambda_function.zip)'
-    Type: String
+  AllowedOrigins:
+    Description: 'Comma-separated list of domains (e.g. example.com) allowed by browser CORS policy, or * for all origins.'
+    Type: List<String>
+    Default: "*"
 
   PublicHostname:
-    Description: 'The public custom domain name for your CloudFront distribution'
+    Description: 'Optional. Replace *.cloudfront.net in TileJSON with a custom hostname (e.g. example.com). See docs on how this value is cached.'
     Type: String
-    Default: 'None'
 
-# ##########################################################################
-# #  S3 Bucket                                             #
-# ##########################################################################
+Outputs:
+  CloudFrontDistributionUrl:
+    Description: 'URL of the CloudFront distribution'
+    Value: !Sub "https://${CloudFrontDistribution.DomainName}"
+    Export:
+      Name: !Sub "${AWS::StackName}-CloudFrontDistributionURL"
+
+Conditions:
+  IsPublicHostnameProvided:
+    Fn::Not:
+      - Fn::Equals:
+          - Ref: PublicHostname
+          - ''
 Resources:
-  S3Bucket:
-    Type: 'AWS::S3::Bucket'
+  LogGroup:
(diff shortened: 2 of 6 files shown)


serverless/aws/protomaps-template.yaml
```diff
@@ -1,36 +1,39 @@
...
-Description: CloudFormation template to create a protomaps infraestructure to serve tiles.
+Description: Serve Z/X/Y tiles through CloudFront + Lambda from an existing S3 bucket.
 Parameters:
   BucketName:
-    Description: 'The name of the S3 bucket where you will store pmtiles files to be served (must be globally unique)'
+    Description: 'Name of an existing S3 bucket with .pmtiles tilesets. Should be in the same region as your CloudFormation stack.'
     Type: String
+    MinLength: 1
 
-  CodeBucketName:
```
bdon, line 8: For now let's use a single bucket that contains both `lambda_function.zip` and the PMTiles, until we have a automated way to inline the code.

I believe inlining is a better solution because otherwise code updates to `lambda_function.zip` require a key change or versioning, since CloudFormation does not detect changes to S3 assets.

serverless/aws/protomaps-template.yaml
```diff
@@ -1,36 +1,39 @@
...
-    Description: 'The S3 key for the Lambda function code (e.g., lambda_function.zip)'
-    Type: String
+  AllowedOrigins:
+    Description: 'Comma-separated list of domains (e.g. example.com) allowed by browser CORS policy, or * for all origins.'
+    Type: List<String>
+    Default: "*"
 
   PublicHostname:
-    Description: 'The public custom domain name for your CloudFront distribution'
+    Description: 'Optional. Replace *.cloudfront.net in TileJSON with a custom hostname (e.g. example.com). See docs on how this value is cached.'
```
bdon, line 15: See CloudFront function below for explanation on how this works. 

The end result here is that the PublicHostname is optional - if you leave it blank the TileJSON will be filled in with the `*.cloudfront.net`  domain name.

serverless/aws/protomaps-template.yaml
```diff
@@ -1,36 +1,39 @@
...
+    Type: AWS::Logs::LogGroup
     Properties:
-      BucketName: !Ref BucketName
-      PublicAccessBlockConfiguration:
-        BlockPublicAcls: true
-        IgnorePublicAcls: true
-        BlockPublicPolicy: true
-        RestrictPublicBuckets: true
+      LogGroupName: !Sub "/aws/lambda/${AWS::StackName}"
+      RetentionInDays: 7
```
bdon, line 36: The default log output has unlimited retention, which can add up; this template defines a new group with one week retention.

serverless/aws/protomaps-template.yaml
```diff
@@ -43,94 +46,87 @@ Resources:
...
             Version: '2012-10-17'
             Statement:
               - Effect: Allow
                 Action:
-                  - logs:CreateLogGroup
                   - logs:CreateLogStream
                   - logs:PutLogEvents
-                Resource: '*'
-        - PolicyName: S3AccessPolicy
+                Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/${AWS::StackName}:log-stream:*"
```
bdon, line 57: Scope the resource more closely instead of using *

serverless/aws/protomaps-template.yaml
```diff
@@ -43,94 +46,87 @@ Resources:
...
             Statement:
               - Effect: Allow
                 Action: s3:GetObject
                 Resource: !Sub arn:aws:s3:::${BucketName}/*
 
-##########################################################################
-#   Lambda Function                                                      #
-##########################################################################
-
-  ProtomapsLambdaFunction:
```
bdon, line 68: Moved this to the end of the YAML file to make it easier to browse when we inline the code.

serverless/aws/protomaps-template.yaml
```diff
@@ -43,94 +46,87 @@ Resources:
...
       Action: lambda:InvokeFunctionUrl
-      FunctionName: !Ref ProtomapsLambdaFunction
+      FunctionName: !Ref LambdaFunction
       Principal: '*'
       FunctionUrlAuthType: NONE
 
-# ##########################################################################
-# #  CloudFront::Distribution                                              #
-# ##########################################################################
+  ViewerRequestCloudFrontFunction:
```
bdon, line 82: This is necessary in the case where `PublicHostname` parameter is left blank.

By default, the Lambda has no way to know what the CloudFront distribution's domain name is.

One way to do this is to set `PUBLIC_HOSTNAME` to reference the CloudFront resource in the CF template, but this introduces a circular dependency between the Lambda and the Distribution.

Some hacks to remove circular dependencies like listed in https://theburningmonk.com/2022/05/how-to-work-around-cloudformation-circular-dependencies/ such as creating the Lambda function ARN by hand don't work; the Lambda Function URL checks on provisioning the referenced function exists.

So instead, what we do is we check for the `PublicHostname` parameter, if blank we install a CloudFront Function that simply attaches the distribution domain name as the `x-distribution-domain-name` header that the Lambda can then read from.

bdon: (The code for reading `x-distribution-domain-name` is not yet in this PR)

bdon: The downside of adding CF Functions is that it costs $0.1 per million requests that go into CloudFront (very little)

serverless/aws/protomaps-template.yaml
```diff
@@ -43,94 +46,87 @@ Resources:
...
+            DomainName: !Select [2, !Split ["/", !GetAtt LambdaFunctionUrl.FunctionUrl]]
             CustomOriginConfig:
               OriginProtocolPolicy: https-only
         DefaultCacheBehavior:
-          TargetOriginId: ProtomapsLambdaOrigin
+          TargetOriginId: LambdaOrigin
           ViewerProtocolPolicy: redirect-to-https
           CachePolicyId: !Ref CachePolicyId
           ResponseHeadersPolicyId: !Ref ResponseHeadersPolicyId
+          OriginRequestPolicyId: b689b0a8-53d0-40ab-baf2-68738e2966ac
```
bdon, line 112: This is hardcoded to `AllViewerExceptHostHeader` as described in the docs:

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-origin-request-policies.html

bdon:
Here is the full output YAML with inlined Lambda code, 494 lines after minification:

https://gist.github.com/bdon/ce3180498561fb0904b7d7ac6df441de

serverless/aws/protomaps-template.yaml
```diff
@@ -43,94 +46,87 @@ Resources:
...
+          FunctionAssociations:
+            Fn::If:
+              - IsPublicHostnameProvided
+              - !Ref "AWS::NoValue"
+              -
```
charliemcgrady, line 117: I think you have an extra line here

serverless/aws/src/index.ts
```diff
@@ -160,20 +160,12 @@ export const handlerRaw = async (
...
-          headers
-        );
-      }
       headers["Content-Type"] = "application/json";
 
       const t = tileJSON(
         header,
         await p.getMetadata(),
-        process.env.PUBLIC_HOSTNAME,
+        process.env.PUBLIC_HOSTNAME || event.headers['x-distribution-domain-name'] || "",
```
charliemcgrady, line 168: Prior to this change, this function was returning a 501 if no  public hostname could be resolved. Here, you're returning an empty string.

Besides notifying a user their setup may not be configured correctly, are there any other reason to avoid returning an empty string?

serverless/aws/package.json
```diff
@@ -7,12 +7,13 @@
...
     "tsx": "^4.7.0",
-    "typescript": "^4.8.4"
+    "typescript": "^4.8.4",
+    "yaml-cfn": "^0.3.2"
   },
   "private": true,
   "scripts": {
     "tsc": "tsc --noEmit --watch",
-    "build": "esbuild src/index.ts --target=es2020 --outfile=dist/index.mjs --format=esm --bundle --platform=node --target=node18 --external:@aws-sdk/client-s3 --external:@aws-sdk/node-http-handler --banner:js=//$(git describe --always) && cd dist && zip lambda_function.zip index.mjs",
+    "build": "esbuild src/index.ts --target=es2020 --minify --outfile=dist/index.js --format=cjs --bundle --platform=node --target=node20 --external:@aws-sdk/client-s3 --external:@aws-sdk/node-http-handler --banner:js=//sha:$(git describe --always)",
```
charliemcgrady, line 16: Any thoughts on including the template as part of the distribution for PMTiles?

It would be great to be able load the compiled yaml directly in JS ('import { pmtilesCFN } from "pmtiles"'). This could be used alongside CfnInclude to embed this infra in existing CDK stacks. Otherwise, clients will need to manually copy the template.

charliemcgrady (approved):

serverless/aws/src/index.ts
bdon: It should probably go back to giving a 501 if both the PUBLIC_HOSTNAME and the distribution domain name header are both blank.

serverless/aws/package.json
bdon: do you mean as part of the `pmtiles` NPM package, that can be included as a AWS CDK dependency?

Originally I kept the Cloudflare and AWS specific code out of the main JS package to reduce versioning churn and dependencies, but since those are stabilizing we can consider this in a next major version.

serverless/aws/protomaps-template.yaml
bdon: This is required since the value to `FunctionAssociations` must be an Array instead of an Object: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-defaultcachebehavior.html#cfn-cloudfront-distribution-defaultcachebehavior-functionassociations

serverless/aws/package.json
charliemcgrady: Yup, any NPM package can be used in typescript based CDK projects.

bdon:
Build output URL: https://pmtiles.io/cloudformation-stack.yaml

PR #446, opened 2024-09-11
Update adapters.ts import FetchSource and add custom headers

reyemtm:
- [ ] Please do not open PRs for new features if you're a first time contributor; first open an issue to confirm that the new feature is appropriate.

### DIFF
--- a/js/adapters.ts
+++ b/js/adapters.ts
@@ -186,7 +186,14 @@ export class Protocol {
       const pmtilesUrl = params.url.substr(10);
       let instance = this.tiles.get(pmtilesUrl);
       if (!instance) {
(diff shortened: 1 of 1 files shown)


reyemtm:
If you are willing to add this let me know and I'll fix the TS errors.

bdon:
This could lead to unexpected behavior because if `transformRequest` is per-request: it could depend on some JS variables outside of the function, or it could be modified after the map is created with `setTransformRequest`. This PR would only capture the output of transformRequest at the first load of the pmtiles archive.

`FetchSource` is explicitly designed for static headers to add that do not change over the lifetime of the object.

reyemtm:
Ok good points. I'll use the fetch source method. Thanks

PR #134, opened 2023-03-12
js: Optimize zxyToTileId performance

hu0p:
This PR improves the performance of `zxyToTileId` by roughly 13-14% via the manual pre-computation and inclusion of the possible `acc` values for zoom. I broke out #132 in a separate PR so that the diff should be easier to digest for this if that's merged first. 

### Benchmarks

Below code requires [benchmark.js](https://www.npmjs.com/package/benchmark).

<details>
 
  <summary>Test Code</summary>

```javascript

const Benchmark = require("benchmark");
var suite = new Benchmark.Suite();

function rotate(n, xy, rx, ry) {
  if (ry == 0) {
    if (rx == 1) {
      xy[0] = n - 1 - xy[0];
      xy[1] = n - 1 - xy[1];
    }
    const t = xy[0];
    xy[0] = xy[1];
    xy[1] = t;
  }
}

const tzValues = [
  0, 1, 5, 21, 85, 341, 1365, 5461, 21845, 87381, 349525, 1398101, 5592405,
  22369621, 89478485, 357913941, 1431655765, 5726623061, 22906492245,
  91625968981, 366503875925, 1466015503701, 5864062014805, 23456248059221,
  93824992236885, 375299968947541, 1501199875790165,
];

function zxyToTileIdNew(z, x, y) {
  if (z > 26) {
    throw Error("Tile zoom level exceeds max safe number limit (26)");
  }
  if (x > Math.pow(2, z) - 1 || y > Math.pow(2, z) - 1) {
    throw Error("tile x/y outside zoom level bounds");
  }

  const acc = tzValues[z];
  const n = Math.pow(2, z);
  let rx = 0;
  let ry = 0;
  let d = 0;
  const xy = [x, y];
  let s = n / 2;
  while (s > 0) {
    rx = (xy[0] & s) > 0 ? 1 : 0;
    ry = (xy[1] & s) > 0 ? 1 : 0;
    d += s * s * ((3 * rx) ^ ry);
    rotate(s, xy, rx, ry);
    s = s / 2;
  }
  return acc + d;
}

function zxyToTileIdOld(z, x, y) {
  if (z > 26) {
    throw Error("Tile zoom level exceeds max safe number limit (26)");
  }
  if (x > Math.pow(2, z) - 1 || y > Math.pow(2, z) - 1) {
    throw Error("tile x/y outside zoom level bounds");
  }

  let acc = 0;
  let tz = 0;
  while (tz < z) {
[...]

PR #470, opened 2024-10-03
use geometric series formula

etienneld:


bdon:
Any description? Can you include the comparison here? https://github.com/protomaps/PMTiles/pull/134

etienneld:
Refactor tzValues calculation to leverage geometric series formula

- Replaced the manual lookup table for `tzValues` with a direct formula based on the properties of a geometric series.
- The formula `(4^z - 1) / (4 - 1)` efficiently computes the same values, eliminating the need for a hardcoded array.
- This approach simplifies code maintenance and avoids memory overhead from lookups, as the formula is computationally inexpensive even for large `z`.

https://www.wolframalpha.com/input?i=Sum%5B2%5Ei+*+2%5Ei%2C+%7Bi%2C+0%2C+n+-+1%7D%5D

etienneld:
This change focuses on maintainability rather than performance optimization.

bdon:
I've merged #527 with results here: https://github.com/protomaps/PMTiles/pull/527#issuecomment-2663128667

PR #498, opened 2024-11-22
fix: use named export for better typing

Kanahiro:
Hi,

I use `pmtiles` module in TypeScript code like this and faced an error.

```typescript
import { Protocol } from 'pmtiles'; // Module '"pmtiles"' has no exported member 'Protocol'.
const protocol = new Protocol();
```

I guess `export * from "./adapters";` caused this and then re-write this with named export.
If it is okay, please merge this.

### DIFF
--- a/js/index.ts
+++ b/js/index.ts
@@ -1,6 +1,6 @@
 import { decompressSync } from "fflate";
 import v2 from "./v2";
-export * from "./adapters";
+export { Protocol, leafletRasterLayer } from "./adapters";
 
 /** @hidden */
 export interface BufferPosition {


bdon:
Did you fix this? It's possible this is related to CJS packages (vs ESM) in your bundler, we need to make a new version that has proper CJS/ESM support.

bdon:
See https://github.com/protomaps/PMTiles/pull/499

Kanahiro:
@bdon I tried `pmtiles-js-v4` and it seems to work fine!

bdon:
I published v4.0.1. Please see if that fixes the issue or creates any new problem.

PR #552, opened 2025-04-10
[Serverless / Cloudflare R2] Use the ETag of the R2 object as a cache busting URL parameter

keichan34:
This is more of a first draft than a completed PR, but I wanted some feedback on the implementation before cleaning it up and updating documentation. I'm especially interested in whether something like this belongs in the repository at all, and what kind of documentation / configuration would be appropriate for a feature of this kind.

Motivation: In the docs, it's recommended to reset the cache by resetting the cache for the whole domain, which can be a bit tedious sometimes. This implementation uses the ETag of the R2 object as a `?v=...` query string on the tile URLs, so uploading a new PMTiles file will trigger changing the URL in the tiles.json endpoint, which won't be in the cache. I set the default cache-control to something short like 30 seconds in my deploy.

I've set the Cache-Control of the tiles.json endpoint to the environment variable to match the original behavior, but it might make sense to split it out to a different variable??

### DIFF
--- a/serverless/cloudflare/src/index.ts
+++ b/serverless/cloudflare/src/index.ts
@@ -132,10 +132,12 @@ export default {
       cacheableHeaders: Headers,
       status: number
     ) => {
-      cacheableHeaders.set(
-        "Cache-Control",
-        env.CACHE_CONTROL || "public, max-age=86400"
-      );
+      if (!cacheableHeaders.has("Cache-Control")) {
+        cacheableHeaders.set(
+          "Cache-Control",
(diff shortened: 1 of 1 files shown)


bdon:
I'm going to close this PR; the implementations in github.com/protomaps are meant to be references and should have the same functionality across Lambda, Cloudflare, go-pmtiles, etc. There are some discrepancies right now, like how JSON metadata is exposed. Adding this cache-busting feature is really useful for caching tilesets that change often but would be behavior specific to the Cloudflare Workers implementation and TileJSON clients.

PR #560, opened 2025-04-26
Toggle visibility of the label layer for line data

roblabs:
When a PMTile layer has a line and has a property of `name`, the PMTiles viewer will draw label.  

This PR fixes the Bug:  *Label still showing when line layer is turned off.*

--- 

*Current PMTiles.io:  line layer is disabled, but label still shows*

 | 

---

*After this PR:  when disabling a line layer via checkbox, both the line & label properly have no `visibility`*


bdon:
Thanks!

PR #577, opened 2025-05-27
app: Ensure points and lines are above polygons in tile view

bdon:
@wipfli


bdon:
There are 3 ways to get to tile view:

1. Enter ZXY on https://pmtiles.io/tile/
2. When inspecting on Map view, click the button in the popup to go to the tile:
CleanShot 2025-06-02 at 11 28 42@2x

3. In Archive view, click the offset of a tile entry:
CleanShot 2025-06-02 at 11 29 20@2x

The archive view is low-level and meant for implementers of the spec and isn't as useful to map data consumers.

PR #578, opened 2025-05-29
Allow to import non minified files from src

simon04:


bdon:
Please provide some context on what problem this solves.

simon04:
Hi, I've added some context to the description (should have done right at the beginning).

bdon:
I don't believe this is a best practice to export both TypeScript in `src` and the transpiled `.js` + `.dts` in `dist`, because we need to make the single build output work for plain Javascript dependent projects.

Can you identify a popular NPM-distributed project that ships all of `.ts` + `.js` + `.dts` that we could refer to?

The 6kb savings is significant, though after compression it's only 2kb. Maybe there is a different way we configure our TypeScript build to get that 2kb savings.

PR #596, opened 2025-10-10
Add MLT tile type

msbarry:
Add tile type for maplibre vector tiles.


bdon:
Merged this and will add to the spec CHANGELOG

PR #597, opened 2025-10-18
Add a type definition for PMTiles header

spider-hand:
Currently `deserialize_header` returns a plain dictionary without explicit field definitions, which makes it hard to know which keys are expected. This PR adds a type definition to enable IDE autocompletion and improve DX.

### DIFF
--- a/python/pmtiles/pmtiles/tile.py
+++ b/python/pmtiles/pmtiles/tile.py
@@ -1,6 +1,7 @@
 import gzip
 import io
 from enum import Enum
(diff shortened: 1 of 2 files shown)


bdon:
Thanks, I uploaded v3.5.0 to PyPI.

larsmaxfield:
3.8 is all good for us. We're at 3.11.

I simply saw the PR and thought to mention it. But I now wonder if there are other imports that would mandate an even newer version?

PR #601, opened 2025-10-24
Add intelligent overview selection for tif to pmtiles conversion

qchenevier:
Hi there,

First, thanks for all the work on the PMTiles format, it's so nice to use.

I noticed that, when using the `rio-pmtiles` package to convert large tiff files to pmtiles, I had performance issues. It took me a while to understand why: when generating low-zoom tiles from large GeoTIFFs, the current `reproject` function of rasterio doesn't load the overviews (see https://rasterio.groups.io/g/main/topic/does_rasterio_warp_reproject/78009234). So it loads "full detail" data on very large extents to compute the overviews: it's very slow and uses lots of memory.

This PR implements a fix by specifying the `OVERVIEW_LEVEL` which needs to be loaded before the `reproject` function call.

Changes:
- CLI: Compute source file's native zoom level and pass to workers (needed to compute the overview level)
- Worker: Compute and use `OVERVIEW_LEVEL` to read appropriate overviews instead of full resolution

Additional small fix:
- Use nodata=255 for RGBA inputs/outputs to prevent warnings by GDAL during conversion like:
  `Warning 1: Value 0 in the source dataset has been changed to 1 in the destination dataset to avoid being treated as NoData. To avoid this, select a different NoData value for the destination dataset.`

> Note: for quick testing you can install it from my fork using:
> ```bash
> pip install -e "git+https://github.com/qchenevier/PMTiles#egg=rio-pmtiles&subdirectory=python/rio-pmtiles"
> ```

### DIFF
--- a/python/rio-pmtiles/rio_pmtiles/scripts/cli.py
+++ b/python/rio-pmtiles/rio_pmtiles/scripts/cli.py
@@ -335,11 +335,12 @@ def pmtiles(
                 warp_options["cutline"] = shapely.wkt.dumps(cutline_rev)
 
         # Resolve the minimum and maximum zoom levels for export.
+        maxzoom_in_file = guess_maxzoom(src.crs, src.bounds, src.width, src.height, tile_size)
         if zoom_levels:
             minzoom, maxzoom = map(int, zoom_levels.split(".."))
         else:
             minzoom = 0
-            maxzoom = guess_maxzoom(src.crs, src.bounds, src.width, src.height, tile_size)
+            maxzoom = maxzoom_in_file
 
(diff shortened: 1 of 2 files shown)


bdon:
This is difficult to unit test without fixtures that have overviews but a user seemed to have success with it: https://sometime.social/notes/aewefko9e3sd010c

> the current reproject function of rasterio doesn't load the overviews

We could remove this once rasterio uses overviews.

bdon:
rio-pmtiles v.1.1.0 released on PyPI, thanks.

bdon:
See https://github.com/protomaps/PMTiles/issues/670 for related fix in `rio-pmtiles` v1.2.2

PR #608, opened 2025-11-21
Properly abort requests using AbortController

### LINKED ISSUE #607: AbortController not respected in all conditions
When using the `tile` and `tilev4` API in JavaScript, the AbortController property is not always respected. In fact, the AbortController is only respected when accessing a tile -- NOT when getting the tile metadata JSON. This can lead to race conditions in dependencies that expect the AbortController to terminate tile metadata requests (e.g., MapLibre's [`loadTileJson`](https://github.com/maplibre/maplibre-gl-js/blob/b8b44cb01f7485964bf6eb7e07b052e90ecb8aea/src/source/load_tilejson.ts#L28C32-L28C39) method).

To resolve this issue, [`abortController.signal.throwIfAborted()`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted) should be called after asynchronous data actions, at least prior to returning a value.

andrewda:
Fixes #607

This PR updates the Protocol.tile/tilev4 methods to properly handle abort signals from the AbortController. Prior to this change, if a user aborted a request using `abortController.abort()`, the abort would only be respected when accessing a tile -- NOT when accessing the TileJSON. This change makes the abort logic consistent throughout the tile methods.

Additionally, Protocol tests were added to cover basic functionality, including aborts for TileJSON and tile requests.

### DIFF
--- a/js/test/adapter.test.ts
+++ b/js/test/adapter.test.ts
@@ -1,12 +1,11 @@
 import assert from "node:assert";
-import { describe, test } from "node:test";
+import { describe, mock, test } from "node:test";
 import { PMTiles, Protocol } from "../src";
-import { MockServer } from "./utils";
-
-const mockserver = new MockServer();
+import { mockServer } from "./utils";
 
 describe("Protocol", () => {
   test("get TileJSON", async () => {
+    mockServer.reset();
     const pmtiles = new PMTiles("http://localhost:1337/example.pmtiles");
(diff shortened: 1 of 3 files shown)


bdon:
Thanks for the PR and additional tests.

I had to rebase off `main` and resolve the merge conflict so I just put this in a separate PR here: https://github.com/protomaps/PMTiles/pull/625

PR #620, opened 2025-12-10
Update `leafletRasterLayer` to dispatch empty image if no raster data

thisisaaronland:
* Update`leafletRasterLayer` to dispatch empty image if `source.getZxy` returns an empty array (no raster data). 
* Run `npm audit fix`

This change means that the `done` callback is always invoked, which is dispatched by the Leaflet.GridLayer `_tileReady` thus ensuring that the layer's "load" event will be fired correctly.

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -14,7 +14,7 @@
       "devDependencies": {
         "@biomejs/biome": "^1.5.3",
         "@types/leaflet": "^1.9.8",
-        "@types/node": "^20.0.0",
+        "@types/node": "^20.19.26",
         "msw": "^2.5.0",
         "tsup": "^8.4.0",
         "tsx": "^4.19.3",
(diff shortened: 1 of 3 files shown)


bdon:
How can I reproduce the original issue?

bdon:
No tool should ever write PNGs with gzip compression applied as it is a waste of CPU cycles. Can you provide `193X.db`?

bdon:
> https://static.sfomuseum.org/aerial/193X.db

This is a SQLite database of PNG images, but the `metadata` table's `format` row is set to `pbf`. 

Instead of doing `gdal2tiles` + `pmtiles-convert` why not use `rio pmtiles` directly?

bdon:
4.3.1 released on https://www.npmjs.com/package/pmtiles

bdon:
Published 4.3.2 with the fix

thisisaaronland:
Confirmed that 4.3.2 works as expected. Thanks!

PR #623, opened 2025-12-15
feat[rio-pmtiles]: allow min and/or max zoom to be omitted when specifying zoom levels

JaffaKetchup:
This means, for example, the user can specify a min zoom level and still leave the max to be auto calculated.

For example:
 - `16..` generates 16-auto
 - `..` generates 0-auto
 - `..16` generates 0-16


bdon:
try setting `PYTHONPATH` to the current directory?

```
PYTHONPATH=. pytest tests
```

bdon:
I uploaded version 1.2.0 to PyPI. https://pypi.org/project/rio-pmtiles/ thank you!

JaffaKetchup:
Great, thanks :)

Issue #631, opened 2026-01-16
Worker optimization


zstadler:
Would you consider replacing a loop with a lookup?

Instead of

https://github.com/protomaps/PMTiles/blob/754e15bf58fa3cd1491bbfd16d48d72a72602596/serverless/cloudflare/src/index.ts#L172-L190

use something like

```typescript
const extToType = {
  mvt: TileType.Mvt,
  pbf: TileType.Mvt, // allow this for now. Eventually we will delete this in favor of .mvt
  png: TileType.Png,
  jpg: TileType.Jpeg,
  webp: TileType.Webp,
  avif: TileType.Avif,
};

const expectedType = extToType[ext];

if (pHeader.tileType !== expectedType) {
  console.debug("fetch returning bad request due to extension mismatch");
  return cacheableResponse(
    `Bad request: requested .${ext} but archive has type ${tileTypeExt(pHeader.tileType)}`,
    cacheableHeaders,
    400
  );
}
```

BTW, similar loop is also part of the AWS worker https://github.com/protomaps/PMTiles/blob/754e15bf58fa3cd1491bbfd16d48d72a72602596/serverless/aws/src/index.ts#L194-L213

PR #641, opened 2026-02-13
Improve serverless type check performance

zstadler:
Resolve https://github.com/protomaps/PMTiles/issues/631


zstadler:
My pleasure!

PR #650, opened 2026-03-24
Fix pmtiles::tileid_to_zxy for platforms where "long" is 32bit

tvoeroes:
`pmtiles::tileid_to_zxy` is incorrect on platforms where "long" is 32bit a.k.a. Windows for tiles with Z >= 16. This PR fixes that issue.


bdon:
Good catch, thank you!

PR #653, opened 2026-04-08
Handle empty MLT tiles the same as MVT

### LINKED ISSUE #652: PMTiles adapter causing crash in MapLibre 5.22.0 when tile does not exist
Original issue is here:  https://github.com/maplibre/maplibre-gl-js/issues/7423

It seems the adapter is explicitly returning an empty Uint8Array for MVT, but not MLT:

https://github.com/protomaps/PMTiles/blob/e90f14cc86fb88cf083f4082d207d1a89eaba320/js/src/adapters.ts#L270

neodescis:
Handle non-existent MLT tiles the same as MVT tiles.

Closes https://github.com/protomaps/PMTiles/issues/652.


neodescis:
I think this should be ready! The test MLT archive is a simple z0 extract.

bdon:
Published v4.4.1, thank you

PR #660, opened 2026-05-13
Add terrain tiles, with elevations encoded as RGB values.

brawer:


spec/v3/spec.md
```diff
@@ -374,6 +374,7 @@ Additionally, this specification defines the following keys, which MAY be includ
 |`attribution`|An attribution to be displayed when the map is shown to a user. Implementations MAY decide to treat this as HTML or literal text. |string|
 |`type`|The type of the tileset |a string with a value of either `overlay` or `baselayer`|
 |`version`|The version number of the tileset|a string containing a valid version according to [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) |
+|`encoding`|Any additional encoding applied to tile data. A value of `terrarium` means the tileset is a Digital Terrain Model with elevation in meters equal to `(red * 256 + green + blue / 256) - 32768`.|string|
```
bdon: Should we clarify that `encoding` only has meaning for raster-type lossless images (PNG, WebP, etc)?

bdon:
The consequences of this spec addition would be:

1) Inspectors like pmtiles.io can render and inspect Elevation tiles with real values, instead of showing the RGB raw tiles
2) tooling GDAL in the future can support PMTiles as an elevation source directly, although this would require changes within GDAL

Seeking additional comments, like if we should support `encoding`:`mapbox`?

spec/v3/spec.md
brawer: Done.

bdon:
Related: https://github.com/mapbox/tilejson-spec/issues/28

bdon:
@ingalls @wipfli per today's conversation, I've merged this into the spec.

jo-chemla:
Nice to hear the pmtiles spec now accepts specifying the `encoding: terrarium` metadata! Is it also made to accept `encoding: terrain-rgb` where elevation from rgb formula is  `elev = -10000 + ((R * 256 * 256 + G * 256 + B) * 0.1)`? Thanks!

bdon:
Can you provide a list of tile generation software and/or map rendering clients that use `encoding: terrain-rgb` right now?

We are encouraging `encoding: "terrarium"` to cover the happy path of using Mapzen Terrain, Mapterhorn and MapLibre together: https://maplibre.org/maplibre-style-spec/sources/#encoding_1

jo-chemla:
The only terrain sources I encountered using TerrainRGB encoding - available on [terrain-viewer](https://terrain-viewer.iconem.com/): 
 - [MapTiler](https://www.maptiler.com/terrain/)
 - of course [Mapbox](https://docs.mapbox.com/data/tilesets/reference/mapbox-terrain-dem-v1/)
 - and then some regional/local terrain DxMs like `ESP - TerrainRGB Nuevo servicio XYZ del Modelo Digital de Elevaciones de España` at [this link](https://xyz-mdt.idee.es/1.0.0/raster-dem/{z}/{x}/{y}.png) or [this one](https://gbank.gsj.jp/seamless/elev/terrainRGB/land/{z}/{y}/{x}.png) from gbank.gsj.jp 
 - also some commercial terrain sources like [Nimbo](https://docs.nimbo.earth/product/terrain-dem/)
 - LINZ was also exposing as TerrainRGB see [mention](https://catalogue.data.govt.nz/dataset/new-zealand-lidar-1m-dsm-survey-index17) but I cannot find it back.

bdon:
Thanks, those seem to be tilesets with RGB encoding but they do not actually define a value in a tileset or renderer specification called `encoding: terrain-rgb` with a precise meaning.

Here is how the current Mapbox GL JS docs define `encoding`: https://docs.mapbox.com/style-spec/reference/sources/#raster-dem-encoding

In the case of tile generation projects like Mapterhorn, we have not observed a need for `encoding: mapbox`, since the sources available in that encoding are proprietary tile APIs and not static tilesets or pmtiles archives. 

If you can provide some examples of places where `encoding: mapbox` is used with static tilesets we can consider adding that to the specification.

jo-chemla:
All the shared TMS/XYZ tilesets share the same underlying TerrainRGB encoding formula from mapbox indeed. When these terrain sources added to terrain-viewer via the BYOD Bring-Your-Own-Data panel, they all display correctly thanks to maplibre terrain-rgb decoding. 

Terrarium encoding has many benefits over terrain-rgb (resolution, finer grained data, going beyond the -10km range etc), so it's great that the ecosystem is pushing towards terrarium encoding within pmiles containers, given that all existing global tilesets listed are proprietary. 

There are some open-data national terrain tiled datasets that do use that mapbox encoding though, like that spanish example. Not sure if this is a living example of a static tileset, but I would assume so - it's also possible this terrain dataset is not only published as terrain-rgb but also as DSM COG. Hope this helps!

> `ESP - TerrainRGB Nuevo servicio XYZ del Modelo Digital de Elevaciones de España` at [this link](https://xyz-mdt.idee.es/1.0.0/raster-dem/%7Bz%7D/%7Bx%7D/%7By%7D.png)

PR #671, opened 2026-09-11
Factor out finalize_header() function

wipfli:
Factors out the the part of the Writer.finalize function which creates the bytes for the header into a separate function called finalize_header().
I would like to use that function to write raster tiles in a single pass in Mapterhorn without having to do the Tempfile extra copy.


bdon:
Uploaded pmtiles 3.8.0 to PyPI

PR #674, opened 2026-09-16
Set mtime in gzip.compress to 0

wipfli:
In Python, gzip.compress includes a timestamp that was only set to default to 0 in version 3.14, see https://docs.python.org/3/library/gzip.html#gzip.compress
As a result when used with older versions, gzip.compress does not produce reproducible output and the output archives have different checksums on multiple runs.


wipfli:
I added a test which fails on main and passes with this fix.

bdon:
Uploaded pmtiles v3.8.1 to PyPI

